PR #5, opened 2024-09-18
aio: cleanup torch model parsing

cryptodeal:
Improve readability parsing pytorch model; optimize for reading slice w/o a single type as tuple.

### DIFF
--- a/zml/aio/torch.zig
+++ b/zml/aio/torch.zig
@@ -379,6 +379,7 @@ pub const PickleData = struct {
                         const key, const val = seq_values[0..2].*;
                         switch (key) {
                             .string => |s| {
+                                // Handle Pytorch specific fields
                                 if (std.mem.eql(u8, s, "_modules") or std.mem.eql(u8, s, "_parameters") or std.mem.eql(u8, s, "_buffers")) {
                                     try self.parseValue(allocator, store, prefix, val);
                                 } else {
--- a/zml/aio/torch.zig
+++ b/zml/aio/torch.zig
@@ -104,7 +104,7 @@ pub const PickleData = struct {
             const args = v.global.args[0].seq[1];
             if (args.len >= 5 and
                 args[0] == .pers_id and
-                args[1] == .int and
+                args[1] == .int64 and
                 args[2] == .seq and args[2].seq[0] == .tuple and
                 args[3] == .seq and args[3].seq[0] == .tuple)
             {
@@ -119,7 +119,7 @@ pub const PickleData = struct {
         var result: [zml.Tensor.MAX_RANK]i64 = undefined;
         for (values, result[0..values.len]) |val, *elem| {
             switch (val) {
-                .int => |int| elem.* = int,
+                .int64 => |int| elem.* = int,
                 else => @panic("Bad value for shape item"),
             }
         }
@@ -175,7 +175,7 @@ pub const PickleData = struct {
             .global => |object| {
                 if (isTensor(v)) {
                     const args = object.args[0].seq[1];
-                    const pidval: *PersId, var offs: u64, const raw_shape: Sequence, const raw_strides: Sequence = .{ args[0].pers_id, @intCast(args[1].int), args[2].seq, args[3].seq };
+                    const pidval: *PersId, var offs: u64, const raw_shape: Sequence, const raw_strides: Sequence = .{ args[0].pers_id, @intCast(args[1].int64), args[2].seq, args[3].seq };
                     const rank = raw_shape[1].len;
                     const shape = dimsFromValues(raw_shape[1]);
                     var strides = dimsFromValues(raw_strides[1]);
@@ -239,7 +239,7 @@ pub const PickleData = struct {
                         allocator.free(key);
                     }
                     const d = try allocator.alloc(i64, size.len);
-                    for (d, 0..) |*di, i| di.* = size[i].int;
+                    for (d, 0..) |*di, i| di.* = size[i].int64;
                     entry.value_ptr.* = .{ .array = .{ .item_type = .int64, .data = std.mem.sliceAsBytes(d) } };
                     return true;
                 } else if (basicTypeCheck(v, "fractions", "Fraction")) {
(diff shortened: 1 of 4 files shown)


zml/aio/torch.zig
```diff
@@ -314,79 +314,76 @@ pub const PickleData = struct {
...
+                            var values: std.ArrayListUnmanaged(switch (tag) {
+                                .int => i64,
+                                .float => f64,
+                                .bool => bool,
+                                else => unreachable,
+                            }) = .{};
+                            try values.append(allocator, val0);
+                            for (seq[1][1..], 1..) |val, i| {
+                                if (std.meta.activeTag(val) != tag) valid_slice = false;
+                                switch (valid_slice) {
```
gwenzek, line 330: not fond of the switch on a boolean.

zml/aio/torch.zig
```diff
@@ -314,79 +314,76 @@ pub const PickleData = struct {
...
-                            else => .boolval,
-                        }, .data = out });
-                    } else {
-                        for (seq[1], 0..) |item, i| {
-                            var new_prefix = prefix;
-                            if (v.isPrimitive()) {
-                                if (prefix.items.len > 0) {
-                                    new_prefix.appendAssumeCapacity('.');
+                            }
+                            const used_tag = switch (tag) {
```
gwenzek, line 342: this feels like an another argument to have less Value types.

zml/aio/torch.zig
```diff
@@ -314,79 +314,76 @@ pub const PickleData = struct {
...
-                                    return val == T;
+                    if (seq[1].len == 0) return;
+                    var valid_slice = true;
+                    switch (seq[1][0]) {
+                        inline .int, .float, .bool => |val0, tag| {
+                            var values: std.ArrayListUnmanaged(switch (tag) {
+                                .int => i64,
+                                .float => f64,
+                                .bool => bool,
+                                else => unreachable,
```
gwenzek, line 325: can you extract the switch to its own line, to clearly separate what is comptime from what is not ?

zml/aio/torch.zig
```diff
@@ -314,79 +314,76 @@ pub const PickleData = struct {
             .pers_id => |pers_id| try self.parseValue(allocator, store, prefix, pers_id.ref),
             .seq => |*seq| switch (seq[0]) {
```
gwenzek, line 315: why are you capture `seq` by reference here ? also I'd rather see `seq` to be destructured into it's type and values, instead of using `seq[0]` and `seq[1]`

zml/aio/torch.zig
```diff
@@ -314,79 +314,76 @@ pub const PickleData = struct {
...
                 },
-                .dict => {
-                    for (seq[1]) |item| {
-                        try self.parseValue(allocator, store, prefix, item);
-                    }
+                .dict => for (seq[1]) |item| {
+                    try self.parseValue(allocator, store, prefix, item);
                 },
                 .kv_tuple => {
                     const key = seq[1][0];
```
gwenzek, line 382: here also you should use tuple destructuring: `const key, cons val = seq[1]`.

zml/aio/torch.zig
```diff
@@ -314,79 +314,76 @@ pub const PickleData = struct {
...
                     const val = seq[1][1];
                     switch (key) {
                         .string => |s| {
-                            if (std.mem.eql(u8, s, "_modules") or std.mem.eql(u8, s, "_parameters") or std.mem.eql(u8, s, "_buffers")) {
+                            if (std.meta.stringToEnum(enum { _modules, _parameters, _buffers }, s)) |_| {
```
gwenzek, line 386: not fond of this stringToEnum hack. the std.mem.eql are fine. You can make a small util at the bottom of the file if you want to shorten it.
also you should document those are pytorch specific fields.

cryptodeal: Fair enough; will revert to `std.mem.eql`

zml/aio/torch.zig
cryptodeal: I think we want to support a diverse set of metadata types though, so not sure sacrificing there makes sense. I've updated the code so the enum values used for the tagged union align with the enum values used in `zml.aio.Value.Slice.ItemType` (e.g. `.bool` becomes `.boolval`, etc.).

zml/aio/torch.zig
cryptodeal: Since it's a slice, I had to destructure as such: `const key, const val = seq_values[0..2].*;`. This is fixed in next commit.

zml/aio/torch.zig
```diff
@@ -312,102 +312,97 @@ pub const PickleData = struct {
...
             .pers_id => |pers_id| try self.parseValue(allocator, store, prefix, pers_id.ref),
-            .seq => |*seq| switch (seq[0]) {
-                .list, .tuple, .set, .frozen_set => {
-                    const elemCheck = struct {
-                        fn call(comptime T: ValueType) fn (v: Value) bool {
-                            return struct {
-                                fn call(val: Value) bool {
-                                    return val == T;
+            .seq => |seq| {
+                const seq_type, const seq_values = seq;
```
gwenzek, line 316: actually why is seq a tuple and not a struct in the first place ? that would make this code easier to read.

zml/aio/torch.zig
```diff
@@ -312,102 +312,97 @@ pub const PickleData = struct {
...
+                switch (seq_type) {
+                    .list, .tuple, .set, .frozen_set => {
+                        if (seq_values.len == 0) return;
+                        var valid_slice = true;
+                        switch (seq_values[0]) {
```
gwenzek, line 321: I think this would gain in clarity if you first checked that seq is a valid slice, then do the parsing.

gwenzek (approved):
I think this is going in the right direction,
but it's just touching the surface of the issues with this implem:

* torch.Values don't need to store allocator. 
* we should not call deinit and walk the graph and free values one by one. Just use arena.
* if you want to keep the deinit function,pass it the allocator, but it doesn't need to be called in ZML.
* I'm not sure we need all those mutable pointers.
* you don't need to dupe key before calling `getOrPut` because the StringHashMap doesn't store the slice, only the hash of its content.

zml/aio/torch.zig
gwenzek: make `isValidSlice` into an helper:

```
fn isValidSlice(seq: Seq) ValueType {
  if (seq.values.len == 0) return true;   
  const t: ValueType = seq.values[0];
  switch (t) {
    .int64, float64, bool => {},
    else => return false,
  }
  for (seq.values) |v| {
    if (v != t) return false;
   }
  return true;
}
```

zml/aio/torch/parser.zig
```diff
@@ -418,16 +418,16 @@ test "Read pickle (simple)" {
                 try testing.expect(key == .string);
```
gwenzek, line 418: did you try "testing.expectEqualsDeep" ? does it work with this struct ?

zml/aio/torch/value.zig
```diff
@@ -100,11 +100,11 @@ pub const ValueType = enum {
     seq,
     string,
     bytes,
-    int,
+    int64,
     bigint,
-    float,
+    float64,
     raw_num,
-    bool,
```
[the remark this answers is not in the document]
gwenzek: but let's rename to boolval here, then do a full s/boolval/bool/g on the full zml

gwenzek (changes requested):
see comments

zml/aio/torch/value.zig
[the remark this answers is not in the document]
cryptodeal: Will do; I'll leave as `boolval `for now, fix RE the remaining feedback, and then in a separate PR to change all instances of `boolval` -> `bool`.

zml/aio/torch.zig
cryptodeal: Yea, that was the previous approach; I can definitely revert, but I do think this is leads to a performance hit. 

Under the previous approach, assuming all values in the slice are of the same type (i.e. the slice is valid), we would iterate over each element twice (once to validate the type and a second time to parse/store the value).

Under this new approach, if all values are of the same type, they are only iterated over once.

Under this new approach, if elements are not all of the same type, there is no performance difference as we end up iterating twice over the uniform elements that precede the element(s) that have a different type, any elements following the non-matching type are only iterated over once. In this case, there is no difference from the previous approach.

IMO, we keep the updated approach as most metadata is likely to be an array of the same type of values, especially large arrays, which is where the old approach would have the most impact on performance.

zml/aio/torch.zig
cryptodeal: Yea, there's no reason this can't be a struct with explicit fields. Will be fixed in the next commit.

zml/aio/torch/parser.zig
cryptodeal: Was able to get that to work, will be fixed in the next commit!

cryptodeal:
Per my conversation with @gwenzek on Discord, I'm planning on handling the additional tasks for torch loader cleanup (namely allocation strategy) in a separate PR.

zml/aio/torch/parser.zig
```diff
@@ -398,64 +399,36 @@ test "Read pickle (simple)" {
...
-                try testing.expect(value.seq[1][0] == .string);
-                try testing.expectEqualStrings("a", value.seq[1][0].string);
-                try testing.expect(value.seq[1][1] == .int);
-                try testing.expect(value.seq[1][1].int == 10);
-            },
-            else => unreachable,
-        }
-    }
+    const expected: []const Value = &.{
+        .{ .seq = .{ .type = .kv_tuple, .values = @constCast(@as([]const Value, &.{ .{ .string = "hello" }, .{ .string = "world" } })) } },
```
gwenzek, line 406: not for this PR, but I'm wondering why values isn't const.
Does the pickle protocol need to modify those when deserializing ?

zml/aio/torch.zig
```diff
@@ -312,102 +312,97 @@ pub const PickleData = struct {
...
+                        var valid_slice = true;
+                        switch (seq.values[0]) {
+                            inline .int64, .float64, .boolval => |val0, tag| {
+                                const ItemType = switch (tag) {
+                                    .int64 => i64,
+                                    .float64 => f64,
+                                    .boolval => bool,
+                                    else => unreachable,
+                                };
+                                var values: std.ArrayListUnmanaged(ItemType) = .{};
```
gwenzek, line 328: can you use ensureCapacityPrecise on values ? that way `toOwnedSlice` isn't needed anymore

zml/aio/torch/parser.zig
gwenzek: also I find it more readable this way.

the `@as([]const Value, &.{...})` can probably be replaced by `&[_]Value{...}`

zml/aio/torch.zig
gwenzek: fair point, let's optimize for the more frequent use case.

gwenzek (approved):
thanks, much easier to read with the `.values` and `.type`

PR #6, opened 2024-09-18
Fix `bazel build //...` and `bazel test //...`

cerisier:
# Description

This PR brings various fixes to allow running `bazel build //...` as well as `bazel test //...`.

# Rationale

Successfully running `bazel build //...` and `bazel test //...` feels like a requirements for any bazel repository.

I can split those into multiple PRs if need be.

# Note

Running `bazel build //...` from the `examples` repo currently fails for 2 reasons:
- Meta-Llama-3.1-8B-Instruct is a gated model and fetching fails with a 403.
- `examples/llama/test.zig` triggers a compile error
```zig
external/zml~/zml/meta.zig:443:9: error: zml.meta.visit is expecting a pointer input to go with following callback signature: fn (void, *buffer.Buffer) void but received: meta.MapType(tensor.Tensor,buffer.Buffer).map(meta.CreateUniqueTuple(1,.{ tensor.Tensor }))
```

### DIFF
--- a/examples/MODULE.bazel
+++ b/examples/MODULE.bazel
@@ -5,9 +5,10 @@ bazel_dep(name = "rules_zig", version = "20240912.0-41bfe84")
 bazel_dep(name = "platforms", version = "0.0.10")
 bazel_dep(name = "zml", version = "0.1.0")
 bazel_dep(name = "aspect_bazel_lib", version = "2.8.1.1")
-
 bazel_dep(name = "rules_oci", version = "2.0.0")
+
 oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
+
 oci.pull(
     name = "distroless_cc_debian12",
     digest = "sha256:b6e1e913f633495eeb80a41e03de1a41aa863e9b19902309b180ffdc4b99db2c",
@@ -17,6 +18,7 @@ oci.pull(
         "linux/arm64/v8",
     ],
 )
+
 use_repo(
     oci,
     "distroless_cc_debian12",
@@ -70,11 +72,12 @@ copy_file(
 """,
     commit = "0bd21da7698eaf29a0d7de3992de8a46ef624add",
     includes = [
-        "stories15M.bin",
         "stories110M.bin",
+        "stories15M.bin",
     ],
     model = "karpathy/tinyllamas",
 )
+
 use_repo(huggingface, "Karpathy-TinyLlama-Stories")
 
 http_file(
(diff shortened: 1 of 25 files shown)


examples/MODULE.bazel
```diff
@@ -5,18 +5,26 @@ bazel_dep(name = "rules_zig", version = "20240912.0-41bfe84")
...
 bazel_dep(name = "aspect_bazel_lib", version = "2.8.1.1")
-
 bazel_dep(name = "rules_oci", version = "2.0.0")
+
 oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
+
 oci.pull(
     name = "distroless_cc_debian12",
-    digest = "sha256:1850aee2ff72864350058d83d681c757d45c885986d15fcca7309b9e5c69f39a",
+    digest = "sha256:b6e1e913f633495eeb80a41e03de1a41aa863e9b19902309b180ffdc4b99db2c",
```
gwenzek, line 14: can you explain what's going on here ?

cerisier: Since I added a platform, the sha changes

cerisier:
Closing for now since the distroless change is not that straightforward and without it, this whole PR is a lie.
I'll have a look at how `rules_oci` could still allow pulling amd64 images on arm64 and will revisit once that's fixed.

examples/MODULE.bazel
gwenzek: what does that mean that an image supports two platform ? if there are two images underlying why isn't there two shas ?

sorry this is all new to me.

gwenzek:
thanks for the exploration.
will think about how to build an ARM64 image.

For buildifier, I don't really mind on the condition it's integrated in the tool chain and I don't have to manually run it.

examples/MODULE.bazel
cerisier: The digest passed to `oci_pull` is actually an identifier that is used to download a `manifest` which can either be:

1. an `index` in case of multi platform, referencing digests for each individual platform images `manifest`.
2. an image `manifest` in case of single platform, referencing digests for its layers.

The new SHA was referencing the `index` manifest containing the digests of the manifests for the various supported platforms (`linux/amd64`, `linux/arm64/v8`)

See:
https://gcr.io/v2/distroless/cc-debian12/manifests/sha256:b6e1e913f633495eeb80a41e03de1a41aa863e9b19902309b180ffdc4b99db2c

The old SHA was referencing the `manifest` of the `linux/amd64` image directly.

Anyway, I wrongly assumed that arm64 was supported and I closed this PR since it would be a bug to use the arm64 image here.

PR #18, opened 2024-09-20
zml/nn: fix resize

gwenzek:
Add more `refAllDecl`

I had to fix resizeBilinear / resizeBicubic.
(85 tests now)

@cryptodeal  can you take over and fix nemo and yaml parsers ?


hugomano (commented):
Thank you! LGTM. I let @cryptodeal doing nemo/yaml parts

PR #25, opened 2024-09-24
Implement scatterSlices

gwenzek:


zml/context.zig
```diff
@@ -121,13 +127,13 @@ pub const Context = struct {
...
-        var platform_: Platform = undefined;
+        var platform_: ?Platform = null;
         var iterator = self.platforms.iterator();
         while (iterator.next()) |entry| {
             if (entry.value.*) |p| {
                 platform_ = p;
             }
         }
-        return platform_;
+        return platform_ orelse @panic("No platform found !");
```
renerocksai, line 137: I nearly panicked over this. But: no platform means: not even a CPU platform -> ZML cannot work -> should be OK-ish to panic. I would prefer an error though and decide myself if I want to panic or send off an alarm to "monitoring" and terminate, etc.

One fictional scenario I can come up with: some code that optionally calls a ZML model, e.g. some web app that server-side gives some LLM hints or whatever. It might want to start and check if it can run inference - and ignore ZML if it can't. In that scenario, a panicking autoplatform() is bad. However, there's a remedy: you can always check the platforms.iterator() yourself and see if you get at least one iteration out of it.

The implication of introducing an error here, is: we'd need to change the examples and tell people. Given how early we are, I wouldn't mind breaking the API for the (in my point of view) better.

Since no platform "should never happen", I'm happy to discuss introducing the error at a later time, in order to not block this PR.

zml/tensor.zig
```diff
@@ -1916,34 +1916,28 @@ pub const Tensor = struct {
...
+        // scoped_log.debug("gatherValues({}, {any}, {})", .{ self, coord_axes, indices });
+        const single_coord, const coord_axes_ = _parseGatherCoord(self, coord_axes);
 
-        meta.assert(val_axes.len > 0, "gatherValues expects 1 or more axes to operate one, received none. Example: `x.gatherValues(.a, indices, .{{}})`", .{});
-        for (val_axes.constSlice(), 0..) |a, i| {
+        meta.assert(coord_axes_.len > 0, "gatherValues expects 1 or more axes to operate one, received none. Example: `x.gatherValues(.a, indices, .{{}})`", .{});
+        for (coord_axes_.constSlice(), 0..) |a, i| {
             if (i > 0) {
-                meta.assert(a == val_axes.get(i - 1) + 1, "gatherValues expects 'axes_' too be sequential. But {any} aren't sequential in {}", .{ axes_, self });
+                meta.assert(a == coord_axes_.get(i - 1) + 1, "gatherValues expects 'coord_axes' too be sequential. But {any} aren't sequential in {}", .{ coord_axes, self });
```
renerocksai, line 1926: too be -> to be

zml/tensor.zig
```diff
@@ -1916,34 +1916,28 @@ pub const Tensor = struct {
...
         for (self._shape.tags(), 0..self.rank()) |t, self_ax| {
-            const maybe_val_ax = std.mem.indexOfScalar(u3, val_axes.constSlice(), @intCast(self_ax));
+            const maybe_val_ax = std.mem.indexOfScalar(u3, coord_axes_.constSlice(), @intCast(self_ax));
             if (indices._shape.hasTag(t)) |id_ax| {
                 // tag is both in self and indices -> it's a batching dim
                 // Note: tags are required for batching.
                 self_kind.appendAssumeCapacity(.batching);
                 indices_batch_axes.appendAssumeCapacity(id_ax);
-                meta.assert(maybe_val_ax == null, "gatherValues expects axes to be either batches or slices axes. Axis {s} has been found both in `axes={any}` and `indices={}`", .{ t, axes_, indices });
+                meta.assert(maybe_val_ax == null, "gatherValues expects axes to be either batches or slices axes. Axis {s} has been found both in `axes={any}` and `indices={}`", .{ t, coord_axes, indices });
```
renerocksai, line 1940: sorry for nitpicking: "or slices axes" is a bit confusing to me. Does it mean "slices of axes" or "sliced axes"?

zml/tensor.zig
```diff
@@ -2238,6 +2232,256 @@ pub const Tensor = struct {
...
+        /// Promise scatter that all coordinates in `indices` are sorted, wrt to the final in memory offset.
+        /// Result is undefined if the promise is violated.
+        indices_are_sorted: bool = false,
+
+        /// Promise scatter that slices don't overlap.
+        /// Result is undefined if the promise is violated.
+        indices_are_unique: bool = false,
+
+        /// Function used to update previous value in `self` with values from `updates`.
+        /// If `update_fn` is not associative (ie the order of execution matter),
```
renerocksai, line 2245: since i'm apparently reviewing typos now, "execution matters".

zml/tensor.zig
```diff
@@ -2238,6 +2232,256 @@ pub const Tensor = struct {
...
+        pub fn override(_: *const anyopaque, old_value: Tensor, new_value: Tensor) Tensor {
+            _ = old_value;
+            return new_value;
+        }
+    };
+
+    /// Update the given tensors, by copying `values` into self slices.
+    /// The slices are chosen at runtime by interpreting indices as coordinates into `self`.
+    /// * `indices` represents a set of coordinates into `self`.
+    ///   For the sake of simplifying the creation of those `indices` tensor,
```
renerocksai, line 2268: tensor -> tensors

renerocksai (approved):
I commented basically on typos and my thoughts about not returning an error in autpPlatform(). Don't want to block; typos can always be fixed in a follow-up if scatter functionality needs to go in quickly.

zml/context.zig
gwenzek: so `zml.contex.init()` already return you an error where there is no platform.
here the panic happen if someone created a `zml.Context` manually, or if someone removed all platforms manually.
I think it's reasonable to panic here.

renerocksai: OK, thanks for that detail. I agree.

PR #26, opened 2024-09-24
zml: sharding

gwenzek:
First rough support for in-process sharding.
The goal of sharding is to allow leveraging several GPUs to collaborate to run one big model.

Current implementation:
* added `platform.sharding()` describing how the set of devices managed by the current process should be used.
* added a `_sharding_info` to shape.
* A tensor / shape axis can be marked as needing sharding using `.withSharding(.{ .a })` or `.withSharding(.{ 1 })`.
* When using `populateModelWeights` tensors with sharded axes, will be populated by sharded buffers.
* when generating the MLIR bytecode for a ZML function the **input** arguments, will see their `_sharding_info` converted into `mhlo.sharding` attributes

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -772,39 +772,47 @@ pub fn ExeWithWeights(comptime func: anytype) type {
         output_buffers: []const [*]*pjrt.Buffer,
 
         /// Internal memory slice used.
-        _data_buffer: []*pjrt.Buffer,
+        _data_buffer: []*anyopaque,
 
         /// And the allocator backing _data_buffer.
         _allocator: std.mem.Allocator,
 
         pub fn initFromModel(allocator: std.mem.Allocator, inner: BaseExe, model: Bufferized(Signature.ModelT)) !Self {
-            const n_buffers = inner.model_buffer_count + inner.args_buffer_count + inner.result_buffer_count;
-
+            const n_input_buffers = inner.model_buffer_count + inner.args_buffer_count;
+            const n_result_buffers = inner.result_buffer_count;
             const n_devices = inner.num_devices;
-            const n_ptrs = (2 + n_buffers) * inner.num_devices;
+            const n_ptrs = (1 + n_input_buffers + 1 + n_result_buffers) * n_devices;
 
-            const data_buffer = try allocator.alloc(*pjrt.Buffer, n_ptrs);
-            errdefer allocator.free(data_buffer);
+            // Allocate once for all the *pjrt.Buffer we need to store,
+            const all_buffers = try allocator.alloc(*anyopaque, n_ptrs);
+            errdefer allocator.free(all_buffers);
 
-            var fba = std.heap.FixedBufferAllocator.init(std.mem.sliceAsBytes(data_buffer));
-            const alloc = fba.allocator();
+            // then split the memory in one slice per device, and also distinguish between input/output.
+            const parts = splitBuffer(
+                *anyopaque,
+                all_buffers,
+                .{ n_devices, n_input_buffers * n_devices, n_devices, n_result_buffers * n_devices },
+            );
+            const input_buffers_per_device: [][*]*pjrt.Buffer = @ptrCast(parts[0]);
+            const all_input_buffers: []*pjrt.Buffer = @ptrCast(parts[1]);
+            const output_buffers_per_device: [][*]*pjrt.Buffer = @ptrCast(parts[2]);
+            const all_output_buffers: []*pjrt.Buffer = @ptrCast(parts[3]);
 
-            const input_buffers = alloc.alloc([*]*pjrt.Buffer, n_devices) catch unreachable;
-            for (input_buffers) |*input| {
-                input.* = (alloc.alloc(*pjrt.Buffer, inner.model_buffer_count + inner.args_buffer_count) catch unreachable).ptr;
+            for (input_buffers_per_device, 0..n_devices) |*input, i| {
+                input.* = all_input_buffers[i * n_input_buffers ..].ptr;
             }
-            fillBuffers(&model, input_buffers, 0, inner.model_buffer_count);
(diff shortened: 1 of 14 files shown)


gwenzek (commented):
I've removed the `postInit` thing to automatically shard linear layers.
let's come back on this later

zml/module.zig
```diff
@@ -565,47 +644,57 @@ fn assignBlockArguments(v: anytype, block: mlir.Block, start: usize) usize {
...
     const LocalContext = struct {
-        index: usize,
-        buffers: []*pjrt.Buffer,
+        index: u32,
+        buffers: []const [*]*pjrt.Buffer,
     };
-    var ctx: LocalContext = .{
-        .index = 0,
+    var capture: LocalContext = .{
+        .index = start,
```
Corendos, line 653: We should try to be consistent on the name of the `LocalContext` variable

zml/module.zig
```diff
@@ -674,56 +765,74 @@ pub fn ExeWithWeights(comptime func: anytype) type {
...
+            const input_buffers = alloc.alloc([*]*pjrt.Buffer, n_devices) catch unreachable;
+            for (input_buffers) |*input| {
+                input.* = (alloc.alloc(*pjrt.Buffer, inner.model_buffer_count + inner.args_buffer_count) catch unreachable).ptr;
+            }
+            fillBuffers(&model, input_buffers, 0, inner.model_buffer_count);
+
+            const output_buffers = alloc.alloc([*]*pjrt.Buffer, n_devices) catch unreachable;
+            for (output_buffers) |*output| {
+                output.* = (alloc.alloc(*pjrt.Buffer, inner.result_buffer_count) catch unreachable).ptr;
+            }
```
Corendos, line 801: Can you make sure we handle these `unreachable` instead ?

Since we are handling the error when allocating `data_buffer`

zml/module.zig
```diff
@@ -875,6 +990,12 @@ fn computeModuleHash(platform: Platform, module: mlir.Module) u64 {
     const api_version = platform.pjrt_api.version();
     writer.writeInt(i64, api_version.major, .little) catch unreachable;
     writer.writeInt(i64, api_version.minor, .little) catch unreachable;
+    const opts = platform.compilation_options;
+    writer.writeByte(if (opts.sharding_enabled) 1 else 0) catch unreachable;
+    writer.writeByte(opts.sharding_axes.len) catch unreachable;
+    for (opts.sharding_axes.constSlice()) |sharding_ax| {
+        writer.print("{s}", .{sharding_ax}) catch unreachable;
+    }
```
Corendos, line 998: Wouldn't this already appear in the generated StableHLO ?

Corendos (commented):
Left some minor comments.

I don't totally understand the logic about `contiguous` in `HostBuffer` though.

From my understanding, we might want to handle two cases:
1) The tensor is not sharded in the model file, so it's already contiguous and we should handle sharding when doing the conversion from `HostBuffer` to `Buffer`.
2) The tensor is already sharded in the model file (or multiple model files) and a `HostBuffer` is a collection of contiguous slices. In that case, each shard gets converted to a `Buffer` naively.

In any case, we currently want to support only 1, so we could move the sharding logic to the place where we convert a `HostBuffer` to a `Buffer` ?

zml/module.zig
gwenzek: agree, but I'm not a big fan of `local_context` and `inner_context`, was wondering if something like `capture` would make it easier to follow to people coming from languages where "capture" is a thing.

zml/module.zig
gwenzek: those are provably unreachable because I've already made the actual allocation line 786.
this is just splitting the pre-existing alloc into smaller slices.
I thought it would be more idiomatic than using pointer arithmetic, but I can also do that if you prefer.

zml/module.zig
gwenzek: oops

zml/pjrtx.zig
```diff
@@ -288,9 +226,9 @@ pub const Event = opaque {
 pub const LoadedExecutable = opaque {
     const inner = InnerMixin(pjrt.LoadedExecutable).inner;
 
-    pub fn deinit(self: *LoadedExecutable, api: *const Api) void {
-        self.inner().deinit(api);
-    }
+    // pub fn deinit(self: *LoadedExecutable, api: *const Api) void {
+    //     self.inner().deinit(api);
+    // }
```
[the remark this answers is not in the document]
gwenzek: because it crashes the Zig compiler, same below. didn't have time to investigate.
the "TODO" below refers to both.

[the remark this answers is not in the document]
gwenzek: (btw I'm tempted to nuke pjrtx.zig)

zml/module.zig
gwenzek: (renamed to context)

zml/module.zig
Corendos: I see. I would probably have used slicing because there is no actual allocations, but that's up to you

zml/module.zig
Corendos: I've seen the `inner_` prefix used many times, that's why I originally used it. But I get your point

zml/module.zig
gwenzek: ok I've rewrote this, let me know what you think

gwenzek:
> I don't totally understand the logic about `contiguous` in `HostBuffer` though.

HostBuffer can have strides, and don't need to be contiguous. This was introduced to support `.pt` where buffer can be stored non-contiguously.
I've also added the `slice1d` utility that allows to slice a `HostBuffer` in several non-contiguous slices.
Since it's possible (and now easy) to create non-contiguous slices, I've added a check in few places that we are indeed working on contiguous `HostBuffer` (which was always the case before adding .pt)

Corendos (approved):

gwenzek:
Thanks for the review and the suggested improvements !

PR #27, opened 2024-09-24
zml: remove dead code

gwenzek:
some dead code or commented out code we should have removed before release.

copyslice is also broken as per the commented out tests.
it will need to be reimplemented if we need it again.


mlir/mlir.zig
```diff
@@ -745,9 +741,9 @@ pub const DictionaryAttribute = struct {
         return NamedAttribute.wrap(c.mlirDictionaryAttrGetElement(self.inner(), @intCast(pos)));
     }
 
-    // pub fn getByName(self: Self, name: [:0]const u8) ?NamedAttribute {
-    //     return NamedAttribute.wrapOr(c.mlirDictionaryAttrGetElementByName(self.inner(), name));
-    // }
+    pub fn getByName(self: Self, name: [:0]const u8) ?NamedAttribute {
+        return NamedAttribute.wrapOr(c.mlirDictionaryAttrGetElementByName(self.inner(), name));
+    }
 };
```
renerocksai, line 747: 👍

renerocksai (approved):
LGTM - trust we don't need Slice & co in hostbuffer.zig

gwenzek:
copySlice was used for testing, but it output wrong results. I think it would be nice to have it restored at some point.

PR #28, opened 2024-09-25
remove bugged Tensor.chunk

gwenzek:
User should use either `chunkExact` or `chunkAllowTrailing` to clarify what they want.

previous `chunk` was crashing on `Tensor(.{.a=10}).chunk(.a, 3)` 
`chunkAllowTrailing` will return **4** tensors: `.{.a=3}, .{.a=3}, .{.a=3}, .{.a=1}`


renerocksai (approved):
LGTM

PR #30, opened 2024-09-30
Tokenizer: hex preprocessing

cryptodeal:
Fix: when merging tokens, `SpecialTokens.unk` length is `1` to prevent index out of bounds.

Feat: Preprocessing support for cases where tokens such as `\n` are not found in vocab, but the hex repr (`<0x0A>`) is.

E.g. For [OpenLlama3b](https://huggingface.co/openlm-research/open_llama_3b/tree/main): 

```py
import torch
from transformers import LlamaTokenizer, LlamaForCausalLM

model_path = 'openlm-research/open_llama_3b'
# model_path = 'openlm-research/open_llama_7b'

tokenizer = LlamaTokenizer.from_pretrained(model_path)

prompt = 'Q: What is the largest animal?\nA:'
print(tokenizer.sp_model.encode(prompt, out_type=str))

input_ids = tokenizer(prompt).input_ids

print(input_ids)
```

The above outputs:
```
['▁Q', ':', '▁What', '▁is', '▁the', '▁largest', '▁animal', '?', '<0x0A>', 'A', ':']
[1, 1059, 31871, 1217, 322, 266, 3682, 6075, 31902, 13, 31849, 31871]
```

Previously, we read this as: `{ 1, 1059, 31871, 1217, 322, 266, 3682, 6075, 31902, 0, 31849, 31871 }`, with the `\n` being read as `SpecialTokens.unk`.

### DIFF
--- a/zml/tokenizer.zig
+++ b/zml/tokenizer.zig
@@ -17,6 +17,7 @@ pub const Tokenizer = struct {
     tokens: [][]const u8,
     token_lookup: std.StringHashMapUnmanaged(u32),
     special_tokens: SpecialTokens,
+    mapped_tokens: std.StringArrayHashMapUnmanaged(u32),
 
     scores: []f32,
     max_token_len: u32,
@@ -66,6 +67,7 @@ pub const Tokenizer = struct {
             .normalizer = normalizer,
             .vocab_size = vocab_size,
             .special_tokens = special_tokens,
+            .mapped_tokens = .{},
         };
     }
 
@@ -120,6 +122,57 @@ pub const Tokenizer = struct {
         return self.token_lookup.get(str);
     }
 
+    fn toHex(dst: []u8, src: []const u8, comptime case: std.fmt.Case) ![]u8 {
+        return std.fmt.bufPrint(dst, "<0x{s}>", .{switch (case) {
+            .lower => std.fmt.fmtSliceHexLower(src),
+            else => std.fmt.fmtSliceHexUpper(src),
+        }});
+    }
+
+    pub fn preprocess(self: *Tokenizer, allocator: std.mem.Allocator, raw: []const u8) ![]const u8 {
+        var out = std.ArrayList(u8).init(allocator);
+
+        var utf8 = (try std.unicode.Utf8View.init(raw)).iterator();
+        while (utf8.nextCodepointSlice()) |codepoint| {
+            // do not handle spaces (normalizer handles)
+            if (std.mem.eql(u8, codepoint, " ")) {
+                try out.appendSlice(codepoint);
+                continue;
+            }
+            // If found in `Tokenizer.lookup`, append it
+            if (self.lookup(codepoint)) |_| {
+                try out.appendSlice(codepoint);
+                continue;
+            }
+
+            const arena = self.arena_state.allocator();
+
+            // If uppercase hex representation is found, append
+            // (e.g. Huggingface models sometimes converts `\n` to `<0x0A>`).
+            var upper_buffer: [100]u8 = undefined;
+            const upper_hex = try toHex(&upper_buffer, codepoint, .upper);
+            if (self.lookup(upper_hex)) |id| {
+                try self.mapped_tokens.put(arena, codepoint, id);
+                try out.appendSlice(upper_hex);
+                continue;
+            }
+
+            // If lowercase hex representation is found, append
+            // (e.g. Huggingface models sometimes converts `\n` to `<0x0A>`).
+            var lower_buffer: [100]u8 = undefined;
+            const lower_hex = try toHex(&lower_buffer, codepoint, .lower);
+            if (self.lookup(lower_hex)) |id| {
+                try self.mapped_tokens.put(arena, codepoint, id);
+                try out.appendSlice(lower_hex);
+                continue;
+            }
+
+            // Otherwise, append the codepoint as is (will become `<unk>`).
+            try out.appendSlice(codepoint);
+        }
+        return out.toOwnedSlice();
+    }
+
     pub const EncodeOptions = struct {
         /// Should the beginning of sentence '<s>' token be added.
         add_bos: bool = true,
@@ -127,9 +180,11 @@ pub const Tokenizer = struct {
         pad_to: u32 = 0,
     };
 
-    pub fn encode(self: *const Tokenizer, allocator: std.mem.Allocator, raw: []const u8, options: EncodeOptions) ![]u32 {
+    pub fn encode(self: *Tokenizer, allocator: std.mem.Allocator, raw: []const u8, options: EncodeOptions) ![]u32 {
         // log.debug("Tokenizer.encode('{s}')", .{raw});
-        const input = if (self.normalizer) |n| try n.normalize(allocator, raw) else raw;
+        const preprocessed_raw = try self.preprocess(allocator, raw);
+        defer allocator.free(preprocessed_raw);
+        const input = if (self.normalizer) |n| try n.normalize(allocator, preprocessed_raw) else preprocessed_raw;
         defer if (self.normalizer) |_| allocator.free(input);
         // log.debug("Tokenizer.encode.normalize -> '{s}'", .{input});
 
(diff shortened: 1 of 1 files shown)


renerocksai:
Tokenizer.encode now expects a non-const "self". --> Llama now needs fixing. I fixed it in #31

renerocksai:
The tokenizer still falls over some prompts:

Model: Llama-3.1-8B-Instruct

```json
{"messages":[{"role":"user","content":"I got the following error, can you please help explain how to fix it?\n\n(base) ➜  zml_private git:(zigtok/vscode) python \"/Users/rs/.vscode/extensions/continue.continue-0.8.
52-darwin-arm64/continue_tutorial.py\"\n/Users/rs/.vscode/extensions/continue.continue-0.8.52-darwin-arm64/continue_tutorial.py:1: SyntaxWarning: invalid escape sequence '\\_'\n  \"\"\"\nTraceback (most recent call last):\n  File \"/Users/
rs/.vscode/extensions/continue.continue-0.8.52-darwin-arm64/continue_tutorial.py\", line 75, in <module>\n    print_sum([\"a\", \"b\", \"c\"])\n  File \"/Users/rs/.vscode/extensions/continue.continue-0.8.52-darwin-arm64/continue_tutorial.p
y\", line 70, in print_sum\n    print(sum(list_to_print))\n          ^^^^^^^^^^^^^^^^^^\nTypeError: unsupported operand type(s) for +: 'int' and 'str'"}],"model":"codellama-7b","max_tokens":1024,"stream":true}
```

The prompt is in messages[0].content

It errors at:

```
thread 352160 panic: index out of bounds: index 4294967295, len 128256
/home/steeve/.cache/bazel/_bazel_steeve/0c5aee2141c34b6ccdebaaf5c5d3b6d5/sandbox/processwrapper-sandbox/1/execroot/_main/external/zml~/zml/tokenizer.zig:266:53: 0x3397d17 in encode (zigtok_lib)
/home/steeve/.cache/bazel/_bazel_steeve/0c5aee2141c34b6ccdebaaf5c5d3b6d5/sandbox/processwrapper-sandbox/1/execroot/_main/demo/zigtok/src/llama_model.zig:56:48: 0x339adec in init (zigtok_lib)
```

renerocksai:
It's probably because of my prompt

zml/tokenizer.zig
```diff
@@ -120,16 +122,69 @@ pub const Tokenizer = struct {
...
+
     pub const EncodeOptions = struct {
         /// Should the beginning of sentence '<s>' token be added.
         add_bos: bool = true,
         add_eos: bool = false,
         pad_to: u32 = 0,
     };
 
-    pub fn encode(self: *const Tokenizer, allocator: std.mem.Allocator, raw: []const u8, options: EncodeOptions) ![]u32 {
+    pub fn encode(self: *Tokenizer, allocator: std.mem.Allocator, raw: []const u8, options: EncodeOptions) ![]u32 {
```
gwenzek, line 183: `encode` should not modify the tokenizer, please keep this pointer const.

zml/tokenizer.zig
```diff
@@ -120,16 +122,69 @@ pub const Tokenizer = struct {
         return self.token_lookup.get(str);
     }
 
+    fn toHex(dst: []u8, src: []const u8, comptime case: std.fmt.Case) ![]u8 {
+        return std.fmt.bufPrint(dst, "<0x{s}>", .{switch (case) {
+            .lower => std.fmt.fmtSliceHexLower(src),
+            else => std.fmt.fmtSliceHexUpper(src),
```
gwenzek, line 128: in general switch/else should be avoided. please explicit all cases as much as possible.

zml/tokenizer.zig
```diff
@@ -120,16 +122,69 @@ pub const Tokenizer = struct {
...
+                try out.appendSlice(upper_hex);
+                continue;
+            }
+
+            // If lowercase hex representation is found, append
+            // (e.g. Huggingface models sometimes converts `\n` to `<0x0A>`).
+            var lower_buffer: [100]u8 = undefined;
+            const lower_hex = try toHex(&lower_buffer, codepoint, .lower);
+            if (self.lookup(lower_hex)) |id| {
+                try self.mapped_tokens.put(arena, codepoint, id);
```
gwenzek, line 165: it's not the right time to do this.
I think that when creating the tokenizer we should convert `<0x0A>` to EOL.
then you wouldn't need to preprocess them and post-process them too.

zml/tokenizer.zig
```diff
@@ -120,16 +122,69 @@ pub const Tokenizer = struct {
...
         add_bos: bool = true,
         add_eos: bool = false,
         pad_to: u32 = 0,
     };
 
-    pub fn encode(self: *const Tokenizer, allocator: std.mem.Allocator, raw: []const u8, options: EncodeOptions) ![]u32 {
+    pub fn encode(self: *Tokenizer, allocator: std.mem.Allocator, raw: []const u8, options: EncodeOptions) ![]u32 {
         // log.debug("Tokenizer.encode('{s}')", .{raw});
-        const input = if (self.normalizer) |n| try n.normalize(allocator, raw) else raw;
+        const preprocessed_raw = try self.preprocess(allocator, raw);
```
gwenzek, line 185: the normalizer should work first, and I'm not sure we need a preprocess at all.

zml/tokenizer.zig
```diff
@@ -144,7 +199,18 @@ pub const Tokenizer = struct {
         var num_tokens: usize = 0;
         var off: usize = 0;
         while (off < input.len) {
-            const utf_len = try std.unicode.utf8ByteSequenceLength(input[off]);
+            var utf_len: usize = try std.unicode.utf8ByteSequenceLength(input[off]);
+            if (utf_len == 1 and input[off] == '<') {
```
gwenzek, line 203: what is this code doing ?

also it's not a good idea to always preprocess text like this. Because the user query should not be allowed to introduce special tokens.

zml/tokenizer.zig
```diff
@@ -206,7 +272,7 @@ pub const Tokenizer = struct {
 
                             meta.assert(std.mem.eql(u8, cur_tok, input[input_off..][0..cur_tok.len]), "current token '{s}' not found in input string '{s}' !", .{ cur_tok, input[input_off..] });
                         }
-                        const concat_tokens = input[input_off..][0 .. cur_tok.len + next_tok.len];
+                        const concat_tokens = input[input_off..][0 .. cur_tok.len + if (tok_buff[i + 1] == self.special_tokens.unk) 1 else next_tok.len];
```
gwenzek, line 275: this looks weird. What is it doing ?

zml/tokenizer.zig
```diff
@@ -306,6 +372,11 @@ pub const Tokenizer = struct {
                 if (output.items.len > 0) try output.append(' ');
             }
 
+            var tmp_buffer: [100]u8 = undefined;
```
gwenzek, line 375: tmp_buff is way too big, but I don't think it's needed if you properly initialize the tokenizer

gwenzek (changes requested):

zml/tokenizer.zig
cryptodeal: Will do

zml/tokenizer.zig
cryptodeal: So the normalizer does work, but it's not producing the same results as the python normalizer. For [OpenLlama3b](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/discussions/35), when testing the snippet on the README, just calling the tokenizer's normalizer on the example prompt, `Q: What is the largest animal?\nA`, yields the following: `['▁Q', ':', '▁What', '▁is', '▁the', '▁largest', '▁animal', '?', '<0x0A>', 'A', ':']`.

This is inconsistent with the behavior of the current Tokenizer implementation.

zml/tokenizer.zig
cryptodeal: This code is checking whether the token is potentially a hex value (e.g. `<0x0A>`) and reads ahead to prevent this token from being split/merged.

zml/tokenizer.zig
cryptodeal: Previously if an `.unk` was being read near the end of the text being encoded, it would result in an index out of bounds as we would attempt to read `[0..cur_tok.len + next_tok.len]`, where `next_tok.len` is `"<unk>".len`, whereas the real character that was unknown only has a length of 1.

gwenzek: ok. but then if next token is `<unk>` we shouldn't merge at all.

PR #31, opened 2024-09-30
FixUp for Llama after PR #30 Tokenizer hex processing 

renerocksai:
Tokenizer.encode now expects non-const self. Llama needed fixing.


gwenzek:
sorry, but the tokenizer shouldn't be mutated while tokenizing. Necessary preparation of the tokenizer should happen before hand. I've also made the feedback on #30, which won't be merged like this.

PR #35, opened 2024-09-30
Cleaned dialect related imports

Corendos:
We were creating useless struct to reexpose `dialect.stablehlo`.

### DIFF
--- a/zml/pjrtx.zig
+++ b/zml/pjrtx.zig
@@ -2,6 +2,7 @@ const builtin = @import("builtin");
 const std = @import("std");
 
 const mlir = @import("mlir");
+const dialect = @import("mlir/dialects");
 const pjrt = @import("pjrt");
 const dtype = @import("dtype.zig");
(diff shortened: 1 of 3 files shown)


zml/pjrtx.zig
```diff
@@ -2,6 +2,7 @@ const builtin = @import("builtin");
 const std = @import("std");
 
 const mlir = @import("mlir");
```
gwenzek, line 4: this import is not needed for `pjrtx.zig`

gwenzek (changes requested):

zml/pjrtx.zig
Corendos: Addressed in https://github.com/zml/zml/pull/35/commits/860c9526ea529bcc7fe2d29c73668d66134b868f

gwenzek (commented):
Good for me, but please get approval from @Steeve, I think he may care about this.

PR #36, opened 2024-09-30
Tokenizer: Handle ByteFallback in Tokenizer

cryptodeal:
Intuitive support for ByteFallback (`\n` -> `<0x0A>` assuming `\n` is not in vocab set).

### DIFF
--- a/examples/llama/main.zig
+++ b/examples/llama/main.zig
@@ -256,8 +256,11 @@ pub fn asyncMain() !void {
     const compile_end = std.time.milliTimestamp();
     log.info("✅ Compiled model in {d} milliseconds! \n", .{compile_end - compile_start});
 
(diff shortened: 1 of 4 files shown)


renerocksai:
There is something fishy going on. I thought we were getting rid of `<0xnn>` in the vocab, and instead putting them into the fallback vocab in their binary form.

Yet when I run this with the codegen model, I get this error:

```
error: current token '<0x20>' not found in input string ' user: for i in 10:
```

So, `cur_tok` is `<0x20>` which is invalid. It should be a space, the binary of 0x20.

gwenzek (changes requested):
This still feels wrong. We can merge ahead of time token_lookup and fallback_lookup.

See my proposition: https://github.com/zml/zml/pull/37

PR #37, opened 2024-10-01
zml.tokenizer: proper byte fallback support

gwenzek:
Some tokenizers have bytes encoded in hex like this: "<0x40>".
This break the tokenization algorithm because the input text will contain "@" not "<0x40>",
and if the input contains "<0x40>" it needs to not be treated as a single byte.
So we replace byte fallbacks strings, by their corresponding character.
This enables the normal tokenization algorithm to work.

This PR also implement the main byte fallback logic, that split unknow utf8 codepoint into bytes.

### DIFF
--- a/examples/llama/main.zig
+++ b/examples/llama/main.zig
@@ -256,8 +256,11 @@ pub fn asyncMain() !void {
     const compile_end = std.time.milliTimestamp();
     log.info("✅ Compiled model in {d} milliseconds! \n", .{compile_end - compile_start});
 
-    const prompt = cli_args.prompt orelse "Once upon a time, there was a little girl named Lily.";
+    const prompt = cli_args.prompt orelse "Q: What is the largest animal?\nA:";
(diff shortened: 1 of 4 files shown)


cryptodeal (approved):

cryptodeal:
Reading through the logic looks good, but it actually is not encoding to match up against python:

Using OpenLLama-3B demo with the prompt `Q: What is the largest animal?\nA:` (demo prompt given in the README of the OpenLLama-3B Repo.

Python correct encoding:
```
[1, 1059, 31871, 1217, 322, 266, 3682, 6075, 31902, 13, 31849, 31871]
```

This PR has reverted the behavior of ZML such that  we're not recognizing the `\n` and it's being encoded as `.unk`:
```
{ 1, 1059, 31871, 1217, 322, 266, 3682, 6075, 31902, 0, 31849, 31871 }
```

gwenzek:
good catch, the sentencepiece proto do have a "byte_fallback" setting, so I've added support to read that.

PR #38, opened 2024-10-01
zml: remove pjrtx

gwenzek:
pjrtx.zig was mean to be a ZML specific wrapper over pjrt.zig.
But pjrt.zig in its current form is pretty usable without wrapper,
the extra layer of indirection is IMHO confusing because it mostly transfer the calls directly to pjrt.zig but make some call suspending.

I moved the few helpers left in pjrtx.zig where they are used, ie 

`pjrtx.Buffer.bufferFromHostBuffer`-> in `buffer.Buffer.from`
`pjrtx.Client.compile` -> in `module.compileModuleToPjrtExecutable`
`pjrtx.Event.await` -> `zml.Platform.awaitEvent()`

This also fix a bug in `pjrtx.Event.await` which was blocking for the event before suspending for the event.


steeve:
bold, I like it

zml/platform.zig
```diff
@@ -94,7 +94,40 @@ pub const Platform = struct {
...
         return self.pjrt_client.getProfiler(self.pjrt_api, options);
     }
+
+    /// Suspend the current co-routine while awaiting for a pjrt event to be over.
+    pub fn awaitEvent(self: Platform, event: *pjrt.Event) !void {
+        defer event.deinit(self.pjrt_api);
+        // If we aren't in a coroutine just use the normal blocking api.
+        if (!asynk.inCoro()) {
+            return try event.await_(self.pjrt_api);
+        }
```
Corendos, line 107: I get the motivation, but it feels like implementation details leaking.

I don't really have another suggestion currently though

zml/tensor.zig
Corendos, line 1: I'd argue that this should be in another PR because it's not related to removing `pjrtx.zig`

Corendos (approved):
Except the two comments I left, I like the fact that this PR removes more code than it adds 😁

zml/tensor.zig
gwenzek: I just stumbled by accident on it. But sure. Let me move that to another PR

zml/platform.zig
gwenzek: I could remove it, but I find it simpler to use the blocking api if asynk is not enabled.

zml/tensor.zig
gwenzek: actually I had reworked it because this disabled test is using "pjrt". But it shouldn't I just remove the test cause `test dynamicSlice` is already pretty similar.

PR #42, opened 2024-10-03
zml: move tests next to the tested function

gwenzek:
note: `dynamicSlice1d()` has been completely removed because the `test dynamicSlice` has already very similar test.


hugomano (approved):
Thanks!

PR #45, opened 2024-10-04
workspace: be a little more precise regarding the huggingface token

steeve:
Indicate that it's also the standard location used by the CLI, and add the environment variable, too.


renerocksai (approved):
I looked through it all and have to say: LGTM!

Will run an update of the docs tonight, once this is merged.

renerocksai:
I removed the `echo -n` requirement from the docs.

PR #47, opened 2024-10-08
Try to read normalizer setting from the huggingface json

gwenzek:
We had issues with the CodeLlama tokenizer. https://huggingface.co/codellama/CodeLlama-7b-hf

Looking at the config, it's actually a "sentencepiece" tokenizer, that uses the special "▁" token to represent spaces.

```
"normalizer": {
    "type": "Sequence",
    "normalizers": [
      {
        "type": "Prepend",
        "prepend": "▁"
      },
      {
        "type": "Replace",
        "pattern": {
          "String": " "
        },
        "content": "▁"
      }
    ]
  },
```

I've added a `Normalizer.fromHfJson` that reads this json and tries to convert this style to our options.
This config can express more things than we can, so it's a best effort.
I've also made the magic space token configurable, requiring adding a small buffer in Normalizer to store it.

There is a small change that can have importance, is that now in most cases the setting "remove_extra_whitespaces" will not be set, meaning that user input will be tokenized differently in face of double spaces. I generally don't think this  is a good practice, but it respect HF defaults.

I've also added a `debug` boolean on the tokenizer struct that shows the individual merges of tokens.
It helped me previously when debugging tokenizer issues.
Before you needed to explicitly uncomment some code to use it, now it's behind a runtime boolean.

### DIFF
--- a/zml/aio/gguf.zig
+++ b/zml/aio/gguf.zig
@@ -57,14 +57,16 @@ pub fn getGgufTokenizer(self: zml.aio.BufferStore, allocator: std.mem.Allocator)
     };
 
     // default options
-    const gguf_normalizer: zml.tokenizer.Normalizer = .{ .flags = .{
-        .escape_whitespaces = true,
-        .remove_extra_whitespaces = true,
-        .add_dummy_prefix = true,
-        .add_dummy_suffix = false,
-        .lower_case_ascii = false,
-        .split_on_punct_ascii = false,
-    } };
+    const gguf_normalizer: zml.tokenizer.Normalizer = .{
+        .escape_whitespaces = zml.tokenizer.Normalizer.sentencepiece_space,
+        .flags = .{
+            .remove_extra_whitespaces = true,
+            .add_dummy_prefix = true,
+            .add_dummy_suffix = false,
+            .lower_case_ascii = false,
+            .split_on_punct_ascii = false,
+        },
+    };
 
     const extra_tokens: u8 = if (tokenizer_impl == .gpt2) 1 else 0;
     const n_tokens: u32 = @intCast(tokens.len + extra_tokens);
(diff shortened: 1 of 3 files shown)


renerocksai (changes requested):
See comment above. Not creating the `1` token at the beginning, will get us even closer to CodeLlama.

gwenzek:
> the tokenizer adds a 1 token at the beginning which is not required.

It's not the tokenizer that adds a 1, it's the caller of the tokenizer.
If you don't want to prepend a "beginning of sentence" token use the `EncodeOptions` accordingly:

```
tokenizer.encode(middle_of_text, allocator, .{ .add_bos = false });
```

Note that this behavior match the default of the huggingface tokenizer.

gwenzek:
> " Hello world": is { 1, 259, 10994, 3186 }, should be: {   29871,  15043,   3186, },  

Even omitting the leading 1, which is a caller setting, I don't know how they get this one. 
The problem is that " Hello world" has a leading space. If we append an extra space and don't cleanup double spaces (which I think you should do on user input), then we get a string with double space at the beginning.
Problem is that double space has a high priority so it get merged first, preventing further merge with "Hello".

renerocksai:
Thx for the `.add_bos`!!!!!!

> Even omitting the leading 1, which is a caller setting, I don't know how they get this one.
The problem is that " Hello world" has a leading space. If we append an extra space and don't cleanup double spaces (which I think you should do on user input), then we get a string with double space at the beginning.
Problem is that double space has a high priority so it get merged first, preventing further merge with "Hello".

I think, this is where the smiley comes into play, see these resources:


- discussion: https://github.com/ggerganov/llama.cpp/issues/3503
- here is how huggingface does it: `_tokenize()`: https://github.com/huggingface/transformers/blob/5af2c6269672cda01c24ad48fab13f14a3ffb746/src/transformers/models/code_llama/tokenization_code_llama.py#L287C14-L301
- which is used in huggingface `tokenize()` : https://github.com/huggingface/transformers/blob/5af2c6269672cda01c24ad48fab13f14a3ffb746/src/transformers/models/code_llama/tokenization_code_llama.py#L255
- related: https://github.com/meta-llama/codellama/issues/87

renerocksai:
CodeLlama-7b-hf works well. I don't see the 1/2 leading spaces edge case as a showstopper tbh. So I'd approve to unlock the improvements.

renerocksai (approved):

PR #50, opened 2024-10-10
zml.aio.json: fix loading empty array

gwenzek:
safetensor loader was segfaulting when loading metadata with an empty json array.

I also rewrote some of the torch loader code to improve the readability.
Notably every slice in `ops.zig` is now const. I've also realized that our strategy to load Pytorch tensor is not future proof see the warning at: https://pytorch.org/docs/stable/storage.html

we need to support the "UntypedStorage". @cryptodeal can you follow up on that ?

### DIFF
--- a/examples/loader/main.zig
+++ b/examples/loader/main.zig
@@ -60,7 +60,7 @@ pub fn asyncMain() !void {
     while (it.next()) |entry| : (i += 1) {
         const host_buffer = entry.value_ptr.*;
         total_bytes += host_buffer.data.len;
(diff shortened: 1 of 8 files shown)


cryptodeal (approved):

cryptodeal:
Interesting RE the future switch to only support "UntypedStorage", can definitely start looking into that so we have an implementation ready prior to the deprecation of `TypedStorage`.

cryptodeal:
Looks good to me!

PR #53, opened 2024-10-12
Wait for events when calling a module

Corendos:
This PR includes two changes:
* Proper handling of multi-device for sharding
* Actually waiting for events to be signaled

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -849,17 +849,22 @@ pub fn ExeWithWeights(comptime func: anytype) type {
 
         pub fn call(self: Self, args: Bufferized(Signature.ArgsT)) Bufferized(Signature.ReturnT) {
             fillBuffers(&args, self.input_per_device, self.inner.model_buffer_count, self.inner.args_buffer_count);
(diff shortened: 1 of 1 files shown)


zml/module.zig
```diff
@@ -849,18 +849,23 @@ pub fn ExeWithWeights(comptime func: anytype) type {
...
                 .results = self.output_per_device,
-                .events = &event,
+                .events = events[0..sharding.num_partitions],
                 // TODO: this allows to tell a specific buffer shouldn't be donated,
                 // even if it has been marked as "can be donated" during compilation.
                 .non_donatable_input_indices = &.{},
             }) catch unreachable;
 
+            for (events[0..sharding.num_partitions]) |e| {
+                e.await_(self.inner.platform.pjrt_api) catch unreachable;
```
gwenzek, line 866: what are the performance implications of always waiting here ?

gwenzek (approved):
so I think the current behavior is fine on Cuda because there is only one backing cuda stream.
but we aren't sure about other platform.
But let's be conservative and explicitly await. If this show up in profiling, will reconsider.

PR #55, opened 2024-10-16
zml.aio.torch: rename zml.aio.Value -> zml.aio.Metadata

gwenzek:
Part of ongoing improvements to the torch loader

* rename zml.aio.Value -> zml.aio.Metadata
* simplify the types zml.aio.Metadata can have
* pickle `parse` function moved to `pickle.zig`
* add dedicated test case
* fix small bug in parse where `short_binstring` was converted to `bin_string`
* pickle `evaluate` function now returns `[]const zml.aio.torch.Value`

note: I didn't modified the core logic.


cryptodeal (approved):
Looks great; good catch on the bug fix!

PR #58, opened 2024-10-17
Llama weights sharding

Corendos:
This PR introduces sharding attributes to Llama weights to take advantage of Tensor Parallelism


gwenzek (approved):
nice job

PR #59, opened 2024-10-17
zml.aio.torch: more refactor

gwenzek:
* use more complicated test cases with long integers
* add tests for protocol 0. Pickle as evolve a lot since protocol 0 was there, 
and it's unlikely we receive such pickle, I thought it was justified since a big part of the code is related to protocol 0 anyway.
* add tests to the "zipped pickle"
* simplify structs hierarchy -> `PickleData` has been merged with `Parser` into `torch.File`.
* rename `torch/value.zig` -> `torch/py_object.zig`
* intermediary results: []pickle.Op, []py_object.AnyPy, should not be stored inside objects but passed around function calls.
* we should use buffered reader to parse the pickle and the zip headers
* python Dict should not  be represented by a singleton seq of seq of seq of kv, just a single seq with keys and values following each other is enough (like in pickle protocol) 
* kwargs should not be mixed with args
* added harder pytorch test with a raw tensor and a `torch.nn.Conv2d` module
* simplified a bit lifecylcle of `BufferStore` by passing the files during init.


cryptodeal (approved):
Looks good; I like the reorganization/cleanup done here a lot!

PR #61, opened 2024-10-18
zml: concatenate tag support

gwenzek:
* add support for `Tensor.concatenate(&.{x, y}, .d)`
* start removing some `broadcastLeft` usage
* use f32 to compute head scaling constant in transformer. We should always compute constant with max precision.

### DIFF
--- a/zml/nn.zig
+++ b/zml/nn.zig
@@ -91,9 +91,9 @@ pub const LayerNorm = struct {
 
     pub fn forward(self: LayerNorm, x: Tensor) Tensor {
         const normed = normalizeVariance(x, self.eps);
-
-        var out = normed.mul(self.weight.broadcastLeft(x.shape()));
-        if (self.bias) |bias| out = out.add(bias.broadcastLeft(x.shape()));
+        const ax = x.axis(-1);
+        var out = normed.mul(self.weight.broadcast(x.shape(), &.{ax}));
+        if (self.bias) |bias| out = out.add(bias.broadcast(x.shape(), &.{ax}));
 
         return out;
     }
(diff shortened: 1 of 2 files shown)


zml/nn.zig
```diff
@@ -91,9 +91,9 @@ pub const LayerNorm = struct {
 
     pub fn forward(self: LayerNorm, x: Tensor) Tensor {
         const normed = normalizeVariance(x, self.eps);
-
-        var out = normed.mul(self.weight.broadcastLeft(x.shape()));
-        if (self.bias) |bias| out = out.add(bias.broadcastLeft(x.shape()));
+        const ax = x.axis(-1);
+        var out = normed.mul(self.weight.broadcast(x.shape(), &.{ax}));
+        if (self.bias) |bias| out = out.add(bias.broadcast(x.shape(), &.{ax}));
```
gwenzek, line 96: I think this less ambiguous than broadcastLeft/Right

zml/nn.zig
```diff
@@ -760,8 +760,8 @@ pub fn sdpa(q_: Tensor, k_: Tensor, v_: Tensor, opts: SdpaOpts) Tensor {
     const dims = helpers.collectDims(.{ .h, .q, .k, .hd }, &.{ q, k, v, attn_mask }, .strict) catch {
         meta.panic(err_template ++ "Inputs have incompatible shapes.", err_args);
     };
-    const sqrtHeadDim: f16 = 1.0 / std.math.sqrt(@as(f16, @floatFromInt(dims.hd)));
-    const scale_logit = if (opts.scale) |s| s else Tensor.scalar(sqrtHeadDim, .f16);
+    const sqrtHeadDim: f32 = 1.0 / std.math.sqrt(@as(f32, @floatFromInt(dims.hd)));
```
gwenzek, line 763: not sure why this is using f16. Avoid unecessary precision loss when computing in f32

renerocksai: agreed

zml/tensor.zig
```diff
@@ -1413,7 +1413,6 @@ pub const Tensor = struct {
     }
 
     pub const Slice = struct {
-        single: ?i64 = null,
         start: i64 = 0,
```
gwenzek, line 1416: the field exists, but there is no downstream support. removing for clarity

renerocksai (approved):
top!

renerocksai:
i'll leave pressing the button to you :smile:

PR #64, opened 2024-10-23
zml.tensor: add cumulativeSum, small refactor of maxPoolND

gwenzek:
This PR adds the "cumulative sum" operator.

This is based on what jax does, which is leveraging reduce window operator,
where for each n in [0..N] you compute the some of elements [0..n] 
I'm not sure how this get optimized in complex programs, to be investigated.

**BREAKING CHANGES**:
* `reduceWindow` is still pretty hard to used but I merged the `padding_shape` and `padding_value` into one operator and made the signature more explicit.
* this also lead me to change a bit how `maxPool1D` and `maxPool2D` take arguments, using proper tuples instead of slices that need to be of specific len.
* `pad` now take options using 
     * "tagged syntax":`x.pad(0, .{ .a = .{ .low = 1, .high = -1 }});` 
     * or the "AOS syntax":  `x.pad(0, &.{.{}, { .low = 1, .high = -1 }});`
     * the "SOA syntax" is **REMOVED**: `x.pad(0, .{ .low = &.{1}, .high = &.{-1} });`

### DIFF
--- a/zml/tensor.zig
+++ b/zml/tensor.zig
@@ -1316,7 +1316,7 @@ pub const Tensor = struct {
         else
             .{ self.dim(a) - 1, 0 };
 
-        return ops.reduceWindow(
+        var res = ops.reduceWindow(
             Tensor.add,
             self,
             Tensor.scalar(0, self.dtype()),
@@ -1328,6 +1328,8 @@ pub const Tensor = struct {
                 .padding = padding[0..rk],
             },
         );
+        res._shape = self._shape;
+        return res;
     }
 
     test cumulativeSum {
--- a/zml/tensor.zig
+++ b/zml/tensor.zig
@@ -1289,10 +1289,21 @@ pub const Tensor = struct {
         return self.sum(axis_).divByConst(self.dim(axis_));
     }
 
-    /// Returns a Tensor containing the cumulative sum of elementes over the given axis.
+    pub const CumulativeSumOpt = struct {
+        start_with_zero: bool = false,
+    };
+
+    /// Returns a Tensor containing the cumulative sum of elements over the given axis.
     /// Ouput shape is the same as input shape.
-    /// [0, 1, 0, 1, 0, 0, 1, 1].cumulativeSum(0) -> [0, 1, 1, 2, 2, 2, 3, 4]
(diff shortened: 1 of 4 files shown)


zml/tensor.zig
```diff
@@ -1273,6 +1273,7 @@ pub const Tensor = struct {
     }
 
     /// Returns a Tensor containing the sum of elements over the given axis.
+    /// Ouput shape is the input shape with the axis_ dim set to 1.
```
Corendos, line 1276:
```suggestion
    /// Output shape is the input shape with the axis_ dim set to 1.
```

zml/tensor.zig
```diff
@@ -1288,10 +1289,59 @@ pub const Tensor = struct {
     }
 
     /// Returns a Tensor containing the mean of elements over the given axis.
+    /// Ouput shape is the input shape with the axis_ dim set to 1.
```
Corendos, line 1292:
```suggestion
    /// Output shape is the input shape with the axis_ dim set to 1.
```

zml/tensor.zig
```diff
@@ -1288,10 +1289,59 @@ pub const Tensor = struct {
...
     pub fn mean(self: Tensor, axis_: anytype) Tensor {
         return self.sum(axis_).divByConst(self.dim(axis_));
     }
 
+    /// Returns a Tensor containing the cumulative sum of elementes over the given axis.
+    /// Ouput shape is the same as input shape.
```
Corendos, line 1298:
```suggestion
    /// Returns a Tensor containing the cumulative sum of elements over the given axis.
    /// Output shape is the same as input shape.
```

Corendos (approved):

gwenzek (commented):
Thanks for the review.
I also added a refactor of the `pad` function. 
This may seem unrelated, but actually it's because they are two ways to compute cumulative sum.

Either `y[i] = sum(k <= i, x[k])` or `y[i] = sum(k < i, x[k])`. 

I thought about adding an option to control which one we want, but it makes `cumulativeSum` harder to use, and you can obtain the second one from the first one using `.pad(0, .{ .a = .{ .low = 1, .high = -1 }});`.

In general I prefer composition of basic primitive over adding flags to each functions.

PR #65, opened 2024-10-23
Fix LSP in Neovim

Corendos:
`zig_exe_path` needs to be set for the LSP to work properly

### DIFF
--- a/examples/.nvim.lua
+++ b/examples/.nvim.lua
@@ -7,6 +7,7 @@ else
         settings = {
             zls = {
                 enable_autofix = true,
+                zig_exe_path = vim.fn.getcwd() .. "/tools/zig.sh"
             },
         },
     }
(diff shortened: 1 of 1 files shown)


renerocksai:
```lua
-- Get the current working directory (CWD)
local cwd = vim.fn.getcwd()

-- Load the local config file if it exists
local f, err = loadfile(".nvim.local.lua")
if not err then
	f()
else
	-- Prepend CWD to relative paths
	local zls_cmd = cwd .. "/tools/zls.sh"
	local zig_exe_path = cwd .. "/tools/zig.sh"

	require("lspconfig")["zls"].setup({
		cmd = { zls_cmd },
		settings = {
			zls = {
				enable_autofix = true,
				zig_exe_path = zig_exe_path,
			},
		},
	})
end
```

why not use the same path concat for the cmd?

Corendos:
You're right, I improved the script in https://github.com/zml/zml/pull/65/commits/29dc891702b385970e1a762524f2307c8e31bfd1

renerocksai (approved):
GOOD!

PR #69, opened 2024-10-25
zml/tensor: use stablehlo.logistic for silu and remove logistic to implement it in sigmoid [BREAKING CHANGE]

hugomano:
For `silu`, `logistic` is more precise than `sigmoid` since the first one is bound to StableHLO while the second one is a custom implementation. 

edit: after the discussion below I removed duplicated `logistic` operator to move the StableHLO call directly in `sigmoid`. The name is more "findable" than `logistic` and more aligned with what we have regarding activations functions; also if the user try to find `logistic` he will find it in the `sigmoid` implementation while the inverse is not true. I added a breaking change mention in the PR title too.

### DIFF
--- a/zml/tensor.zig
+++ b/zml/tensor.zig
@@ -1252,7 +1252,7 @@ pub const Tensor = struct {
     /// silu(x) = x σ(x)
     /// https://paperswithcode.com/method/silu
     pub fn silu(x: Tensor) Tensor {
-        return x.mul(x.sigmoid());
+        return x.mul(x.logistic());
     }
 
     /// Returns a Tensor containing the softmax function applied to each element of the input Tensor.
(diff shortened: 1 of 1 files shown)


renerocksai (approved):
Nice find! Makes sense!

Q: shouldn't we alias Tensor.sigmoid() to Tensor.logistics()?
Then it would be a win for everybody using sigmoid.

gwenzek:
yes, feel like we should remove the duplicate. The reason why we add sigmoid was to try to support metal, but since we dropped metal ...
also if we add metal again we can do `if (self.getContext().platform == .metal)`...

Corendos (approved):

renerocksai:
It's fine with me. Though it would also have been fine and even covered both preferences, if we had made a `pub const logistics = sigmoid` alias.

hugomano:
> It's fine with me. Though it would also have been fine and even covered both preferences, if we had made a `pub const logistic = sigmoid;` alias.

done :)

PR #75, opened 2024-11-04
zml.tensor: add triangular and toDiagonal

gwenzek:
- triangular: zeroes out the upper right part of a matrix. Options control how much is actually zeroes with offset from the diagonal.
- toDiagonal: equivalent to pytorch "diag_embed". The base case is converting a vector into a diagonal matrix. We already have a `diag` operator doing this, but the name was wrong cause in general `diag` does the reverse operation of extracting the diagonal from a matrix into a vector. Also it was using old style ZML, and was a bit too restrictive on input shapes.
- elu: one more activation function. it's getting a bit out of hand, I'm thinking about moving them out of Tensor namespace into zml.nn.Activation wdyt ?

### DIFF
--- a/zml/nn.zig
+++ b/zml/nn.zig
@@ -52,6 +52,7 @@ pub const Activation = union(enum) {
     tanh,
     relu,
     leakyReLU: f32,
+    elu: f32,
     silu,
     gelu,
     quick_gelu,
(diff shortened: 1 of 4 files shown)


Corendos (approved):

Corendos:
> elu: one more activation function. it's getting a bit out of hand, I'm thinking about moving them out of Tensor namespace into zml.nn.Activation wdyt

No strong opinions, if you feel it's better go ahead 😁

PR #77, opened 2024-11-04
async: add threaded implementation

steeve:
This implements the `async` module using a thread pool. This is now the default.

Implementations can be swapped using the `@zml//async:impl=[threaded,zigcoro]` flag.

Because io_uring is not enabled in the default seccomp profile, some container implementations will block on io_uring calls. Use the threaded implementation to work around this, for now.

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -49,6 +49,7 @@ pub const CompilationContext = struct {
     _unique_id: u64 = 10000,
     _tracer: Tracer,
 
+    _previous: ?*CompilationContext = null,
     threadlocal var _current: ?*CompilationContext = null;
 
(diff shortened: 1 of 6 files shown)


zml/module.zig
```diff
@@ -95,12 +96,14 @@ pub const CompilationContext = struct {
     }
 
     pub fn activate(self: *CompilationContext) void {
+        self._previous = _current;
```
gwenzek, line 99: wouldn't it be safer to always spinup a new thread for compilation ?

gwenzek (approved):

zml/module.zig
steeve: It would, but that interface would be at a higher level, not in `CompilationContext`

PR #79, opened 2024-11-06
PJRT: Profiling Output to JSON

cryptodeal:
Implements conversion of PJRT protobuf profiling output to JSON format. Mirrors the XLA implementation, but is much quicker to compile.


pjrt/third_party/re2/re2.cc
```diff
@@ -0,0 +1,28 @@
...
+  bool isJaxOpType(const char *op_type, size_t len) {
+    absl::string_view op_type_view(op_type, len);
+    // Jax op type should start with lowercase character or underscore.
+    // If it contains '[]', it must end with ']' and whatever chars inside
+    // it are considered as a match.
+    static const LazyRE2 kJaxOpTypeRegEx = {"[a-z_][a-z0-9_]*(\\[.*\\])?"};
+    return RE2::FullMatch(op_type_view, *kJaxOpTypeRegEx);
+  }
+
+  bool isNumber(const char *str, size_t len) {
```
gwenzek, line 22: this can be done in zig with std.fmt.parseInt

pjrt/third_party/re2/re2.cc
```diff
@@ -0,0 +1,28 @@
...
+
+extern "C" {
+  bool isTfOpName(const char *op_name, size_t len) {
+    absl::string_view op_name_view(op_name, len);
+    static const LazyRE2 kTfOpNameRegEx = {"[A-Za-z0-9.][A-Za-z0-9_.\\/>-]*"};
+    return RE2::FullMatch(op_name_view, *kTfOpNameRegEx);
+  }
+
+  bool isJaxOpType(const char *op_type, size_t len) {
+    absl::string_view op_type_view(op_type, len);
```
gwenzek, line 14: not a big fan of special casing. Also this is used to strip information ouptutted by the compiler, so I think we don't want to do that.  

overall I think we would be better by not trying to guess the source framework based on the casing of the operator name, and drop this dependency on re2.

pjrt/convert/utils/xplane_visitor.zig
```diff
@@ -0,0 +1,293 @@
...
+        };
+    }
+
+    pub fn forEachStat(
+        self: *const XEventMetadataVisitor,
+        allocator: std.mem.Allocator,
+        cb: fn (allocator: std.mem.Allocator, xstat: XStatVisitor, ctx: ?*anyopaque) std.mem.Allocator.Error!void,
+        ctx: ?*anyopaque,
+    ) !void {
+        for (self.stats_owner.stats.items) |*stat| {
```
gwenzek, line 23: I don't really like this visitor pattern. We use it in zml because we need to handle user defined struct, but here you know all the structs, so I'm not sure what you gain compared to writing the for loop every time.

pjrt/convert/utils/xplane_schema.zig
```diff
@@ -0,0 +1,554 @@
...
+
+pub const kXlaAsyncOpLineName = "Async XLA Ops";
+
+pub const kHostThreadsPlaneName = "/host:CPU";
+pub const kGpuPlanePrefix = "/device:GPU:";
+pub const kTpuPlanePrefix = "/device:TPU:";
+pub const kCustomPlanePrefix = "/device:CUSTOM:";
+
+pub const HostEventType = enum(u16) {
+    kUnknownHostEventType = 0,
```
gwenzek, line 12: those kPascalCase should be replace by snake_case.

alternatively instead of using different name in the enum and in the json,
 just use the human readable name in the enum by using the `@"ExecutorState::Process"` syntax.

pjrt/convert/utils/trace_utils.zig
```diff
@@ -0,0 +1,14 @@
+// Constants used as trace_viewer PID (device_id in trace_events.proto).
+// PID 0 is unused.
+// Support up to 500 accelerator devices.
+pub const kFirstDeviceId = 1;
```
gwenzek, line 4: just inline this inside "trace_container.zig" also no need for the "k" prefix. and use snake_case

pjrt/convert/utils/math_utils.zig
```diff
@@ -0,0 +1,9 @@
+const std = @import("std");
+
+pub fn nanoToPico(n: u128) u128 {
```
gwenzek, line 3: please keep those helpers next to where they are used, or inline them.

gwenzek (commented):
thanks for the work, can you share with me a json file produced by this implementation ?

cryptodeal:
Here's the JSON outputted from `open_llama3b.trace.pb` file you'd sent over for testing:
[open_llama3b.trace.json](https://github.com/user-attachments/files/17653546/open_llama3b.trace.json)

pjrt/convert/utils/xplane_visitor.zig
cryptodeal: Yea, that makes sense; I wrote it this way purely because that's how the XLA implementation handles it, but will rewrite as for loops.

pjrt/third_party/re2/re2.cc
cryptodeal: Looking into this bit; this is based directly on how XLA parses: https://github.com/openxla/xla/blob/cb5bdb601bdb352afaa804b5b4a8826734772e3b/xla/tsl/profiler/utils/tf_op_utils.cc#L129

I agree it would be best to get rid of the `re2` and `abseil-cpp` dependencies.

cryptodeal:
Modified `TraceConverter` so that it accepts a `[]const u8` as input (i.e. can pass encoded profiling protobuffer directly to the converter and only write the JSON to a file).

Removed deps on `re2`/`abseil-cpp` as we're now using `mvzr` for simple regex library. Additionally, no information is being stripped from the resulting JSON while parsing TF/Jax ops, it's simply used to correctly categorize stat types based on operation name at this point (feature parity w XLA).

pjrt/third_party/re2/re2.cc
gwenzek: I understand XLA is doing weird stuff, but I don't think we should reproduce that part. The goal is to parse the files generated while profiling ZML, not tensorflow or jax. Also I believe this way of handling the input is very brittle, because it can categorize ZML operators as tensorflow / jax and apply logic where it's not warranted.
I'd rather remove this logic altogether.

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,223 @@
...
+        args: std.StringHashMapUnmanaged([]const u8) = .{},
+    };
+
+    pub const Device = struct {
+        name: []const u8 = &[_]u8{},
+        device_id: u32 = 0,
+        resources: std.AutoHashMapUnmanaged(u32, Resource) = .{},
+
+        pub fn mutableResource(self: *Device, allocator: std.mem.Allocator, resource_id: u32) !*Resource {
+            const entry = try self.resources.getOrPut(allocator, resource_id);
```
gwenzek, line 53: I don't really like this pattern because it default inits value. I'd rather you use `getOrPutValue` directly from the caller.

pjrt/convert/utils/xplane_visitor.zig
```diff
@@ -0,0 +1,252 @@
...
+            .metadata = plane.getEventMetadata(event.metadata_id),
+            .type_ = plane.getEventType(event.metadata_id),
+        };
+    }
+
+    pub fn hasDisplayName(self: *const XEventVisitor) bool {
+        return self.metadata.display_name != .Empty;
+    }
+
+    pub fn displayName(self: *const XEventVisitor) []const u8 {
```
gwenzek, line 94: you should return an optional here, and drop `hasDisplayName()`

pjrt/convert/trace_converter.zig
```diff
@@ -0,0 +1,178 @@
...
+        var writer = buffer.writer();
+        try writer.writeAll(
+            \\{"displayTimeUnit":"ns","metadata":{"highres-ticks":true},"traceEvents":[
+        );
+
+        // TODO: finish implementing
+        const Entry = std.AutoHashMapUnmanaged(u32, TraceContainer.Device).Entry;
+        const pairs = try allocator.alloc(Entry, self.container.metadata.devices.count());
+        defer allocator.free(pairs);
+        var iter = self.container.metadata.devices.iterator();
```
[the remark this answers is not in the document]
gwenzek: if you really need the index, please convert the HashMap to ArrayHashMap, and then you can directly `std.mem.sort(devices.entries)`,  or call `devices.sort()` if you need to rebuild the index afterward.

pjrt/convert/trace_converter.zig
```diff
@@ -0,0 +1,178 @@
...
+                picoToMicro(event.timestamp_ps),
+                picoToMicro(duration_ps),
+                event.name,
+            });
+            if (event.args.count() != 0) {
+                try writer.writeAll(
+                    \\,"args":{
+                );
+                const ArgsEntry = std.StringHashMapUnmanaged([]const u8).Entry;
+                const sorted_args = try allocator.alloc(ArgsEntry, event.args.count());
```
gwenzek, line 151: same thing, I have a hard time justifying a hashmap here for args.

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        // Trace viewer (non-streaming) has scalability issues, we need to drop
+        // events to avoid loading failure for trace viewer.
+        const viewer_max_events = try getTraceViewerMaxEvents();
+        self.capEvents(viewer_max_events);
+        return self;
+    }
+
+    pub fn mutableDevice(self: *TraceContainer, allocator: std.mem.Allocator, device_id: u32) !*Device {
```
gwenzek, line 193: please also replace by `getOrPutValue`

pjrt/convert/utils/xplane_schema.zig
```diff
@@ -0,0 +1,300 @@
+const std = @import("std");
+const tf_op_utils = @import("tf_op_utils.zig");
+
+pub const HostEventType = enum(u16) {
```
gwenzek, line 4: please add a comment explaining the casing is important to match the contant of the protobuffer

pjrt/convert/utils/xplane_schema.zig
```diff
@@ -0,0 +1,300 @@
...
+            .ParseExampleProduce,
+            .ParseExampleConsume,
+            => true,
+            else => false,
+        };
+    }
+};
+
+pub const StatType = enum(u16) {
+    UnknownStatType = 0,
```
gwenzek, line 132: just use "unknown", the type system knows it's a StateType

gwenzek (changes requested):

pjrt/convert/trace_converter.zig
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn toJson(self: *TraceConverter, allocator: std.mem.Allocator) ![]const u8 {
+        var buffer = std.ArrayList(u8).init(allocator);
+        errdefer buffer.deinit();
+        var writer = buffer.writer();
+        try writer.writeAll(
+            \\{"displayTimeUnit":"ns","metadata":{"highres-ticks":true},"traceEvents":[
+        );
+
+        // TODO: finish implementing
```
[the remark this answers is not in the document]
cryptodeal: Oops, this is just one that I forgot to remove

pjrt/convert/utils/xplane_schema.zig
cryptodeal: When initializing `XPlaneVisitor`, we need to be able to convert from the string representation of the enum value to the actual zig `StatType` enum type. XLA has the string representation for unknown ops as `"UnknownStatType"`. 

For simplicity of the `StatType.fromString` logic, imo it's best to leave as is so we can purely leverage `std.meta.stringToEnum`.

pjrt/convert/utils/xplane_visitor.zig
```diff
@@ -0,0 +1,252 @@
...
+
+    pub fn @"type"(self: *const XStatVisitor) ?xplane_schema.StatType {
+        return self.type_;
+    }
+
+    pub fn name(self: *const XStatVisitor) []const u8 {
+        return self.metadata.name.getSlice();
+    }
+
+    pub fn toString(self: *const XStatVisitor, allocator: std.mem.Allocator) ![]const u8 {
```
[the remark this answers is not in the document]
cryptodeal: I went with `AnyWriter` instead, since that works nicely with the `ArrayList`/`std.fs.File` writers.

pjrt/convert/utils/xplane_schema.zig
gwenzek: oh, didn't realize this could happen in protobuf too, I thought it was your fallback for unhandled protobuffer fields

pjrt/convert/trace_container.zig
```diff
@@ -184,12 +224,17 @@ pub const TraceContainer = struct {
         return self;
     }
 
-    pub fn mutableDevice(self: *TraceContainer, allocator: std.mem.Allocator, device_id: u32) !*Device {
-        const entry = try self.metadata.devices.getOrPut(allocator, device_id);
-        if (!entry.found_existing) {
-            entry.value_ptr.* = .{};
+    pub fn getOrPutDevice(self: *TraceContainer, allocator: std.mem.Allocator, key: u32) !GetOrPutResult(u32, ?trace_events_proto.Device) {
```
gwenzek, line 227: there is a misunderstanding there. This is not what I asked for. 
I want to you to adapt the data structure to the algorithm.

* if you need to insert stuff by key, please use an `AutoArrayHashMap`, and `getOrPutValue`
* if you don't please use an `ArrayList(KV)` and use `.append`. 

The answer to this question probably depend on the thing you are storing. Maybe devices need to use a hash map and arguments an arraylist.

in any case, please explain for each container why you chose this one and not the other.

pjrt/third_party/re2/re2.cc
gwenzek: I don't think this was addressed

gwenzek (changes requested):
there is a misunderstanding there. This is not what I asked for.
I want to you to adapt the data structure to the algorithm.

if you need to insert stuff by key, please use an AutoArrayHashMap, and getOrPutValue
if you don't please use an ArrayList(KV) and use append.

The answer to this question probably depend on how is the data structure in the protobuf, and on how you need to represent it in the json.
If  you need to merge information from different part of the protobuffer then you need an ArrayHashMap, if you don't merge things then ArrayList is enough.

Maybe devices need to use a hash map and arguments an arraylist.
in any case, please explain for each container why you chose this one and not the other.

In general Zig std is pretty high quality. If you find yourself implementing helper around std it means you're doing something wrong. Before you were implementing "sort" on a HashMap, now you are implementing "getOrPutValue" on an ArrayList. This means you're using the wrong container for your algorithm. 

also I don't think my comments about the unwarranted regex logic of "isJax", "isTensorflow" was accounted for, yet the conversation was marked as resolved. Can you be more conservative when resolving discussions ?

pjrt/third_party/re2/re2.cc
cryptodeal: Gotcha, I'll just remove this logic entirely then.

cryptodeal: Sorry, somehow I missed the comment you'd left on this 4 days ago after I'd initially revised to remove the dependencies on `re2`/`abseil-cpp`.

cryptodeal:
> also I don't think my comments about the unwarranted regex logic of "isJax", "isTensorflow" was accounted for, yet the conversation was marked as resolved. Can you be more conservative when resolving discussions ?

Understood, I think I resolved as I'd read your initial comment as needing to remove the dependencies and prevent trimming potentially useful information when filtering op name by string, but I see what you mean now and will completely remove this logic.

pjrt/convert/trace_container.zig
cryptodeal: For any protobuf `map<KeyType, ValueType>`, protobuf encodes as a repeated message where the message type is a pair of `KeyType, ValueType`, which is why `zig-protobuf` handles as `ArrayList(KV)`. However, protobuf does not allow duplicate keys, so a better translation would probably be `AutoArrayHashMap`.

cryptodeal:
Upon further reflection, I believe that `AutoArrayHashMap`/`StringArrayHashMap` are the best data structures for use here; namely, protobuf maps do not allow duplicate key entries, so this ensures deduplication works consistently with the XLA implementation while providing a simple means to sort the entries.

pjrt/convert/trace_container.zig
cryptodeal: For `devices`, I've chosen `AutoArrayHashMap` as it enforces deduplication on the key, which is consistent with the protobuf map type/XLA implementation.

For `args`, I've chosen `StringArrayHashMap` as it enforces deduplication on the key, which is consistent with the protobuf map type/XLA implementation. Additionally, the `KV` pairs here are directly outputted as JSON Key/Value pairs; this ensures the resulting output is valid JSON (no duplicate keys).

For `resources`, I've chosen `AutoArrayHashMap` as it enforces deduplication on the key value, which is consistent with the protobuf map type/XLA implementation.

pjrt/convert/utils/xplane_visitor.zig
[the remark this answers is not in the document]
gwenzek: I think this changes havn't been pushed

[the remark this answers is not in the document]
gwenzek: also can you rewrite this function to directly take an xplane and an xstat instead of a "xstatevisitor" ?

Since this visitor doesn't have state, it's only used to "extend" the XState proto, which is better done by just writing function using plain data as input.

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,211 @@
...
+        args: std.StringArrayHashMapUnmanaged([]const u8) = .{},
+    };
+
+    fn buildDeviceAndResources(allocator: std.mem.Allocator, device_id: u32, plane: *const xplane_visitor.XPlaneVisitor, device: *Device) !void {
+        device.name = plane.name();
+        device.device_id = device_id;
+        const sort_by_ordinal = (device_id == host_threads_device_id);
+        var ordinal: u32 = 0;
+        for (plane.plane.lines.items) |*l| {
+            var line = xplane_visitor.XLineVisitor.init(plane, l);
```
gwenzek, line 67: here instead of creating a "XLineVisitor" then calling displayeId and displayName on it, can you create a `fn xLineDisplayId(plane, line)` and `fn xLineDisplayName(plane, line)` and call that ?

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,211 @@
...
+
+        try buildDeviceAndResources(allocator, device_id, xplane, device_entry.value_ptr);
+
+        // Convert events.
+        for (xplane.plane.lines.items) |*l| {
+            const xline = xplane_visitor.XLineVisitor.init(xplane, l);
+            const resource_id: u32 = @intCast(xline.displayId());
+            if (std.mem.eql(u8, xline.displayName(), xla_async_op_line_name)) continue;
+            for (xline.line.events.items) |*e| {
+                const xevent = xplane_visitor.XEventVisitor.init(xline.plane, xline.line, e);
```
gwenzek, line 95: same here with xline. This code is obfuscated, since "xplane" and "l" have become "xline.plane" and "xline.line". With less intermediary objects it would be easier to follow.

pjrt/convert/utils/xplane_visitor.zig
```diff
@@ -0,0 +1,234 @@
...
+                    try self.stat_type_by_id.put(allocator, metadata_id, stat_type);
+                    try self.stat_metadata_by_type.put(allocator, stat_type, &stat_metadata.value.?);
+                    break;
+                }
+            }
+        }
+    }
+
+    pub fn getEventType(self: *const XPlaneVisitor, event_metadata_id: i64) ?xplane_schema.HostEventType {
+        return self.event_type_by_id.get(event_metadata_id);
```
gwenzek, line 205: instead of returning an "?HostEventType" return HostEventType with .unknown for non found values.

pjrt/convert/utils/xplane_visitor.zig
```diff
@@ -0,0 +1,234 @@
+const std = @import("std");
+const xplane_proto = @import("//tsl:xplane_proto");
+const xplane_schema = @import("xplane_schema.zig");
+
+pub const XEventMetadataVisitor = struct {
```
gwenzek, line 5: this struct is just a tuple, can you remove it ?

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,211 @@
...
+                    if (xstat.type()) |t| {
+                        if (t.isInternalStat()) continue;
+                        if (t == .step_name) {
+                            event.name = stat_str;
+                        }
+                    }
+                    try event.args.put(allocator, xstat.name(), stat_str);
+                }
+
+                for (xevent.event.stats.items) |*s| {
```
gwenzek, line 125: here also there is obfuscation: xevent.event is just our original `e`. I don't think this "renaming"  helps

gwenzek (changes requested):
Thanks, it's improving. The core algorithm looks good now,
 but can you streamline the code by removing all the visitor struct that barely do anything ?

you would need some helper function that work directly on the protobuf object themselves instead of wrapping them in visitor firsts.

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,239 @@
...
+pub const tpu_plane_prefix = "/device:TPU:";
+pub const custom_plane_prefix = "/device:CUSTOM:";
+
+pub const TraceContainer = struct {
+    metadata: Trace = .{},
+    events: std.ArrayListUnmanaged(*TraceEvent) = .{},
+
+    pub const Trace = struct {
+        devices: std.AutoArrayHashMapUnmanaged(u32, Device) = .{},
+        trace_events: std.ArrayListUnmanaged(TraceEvent) = .{},
```
gwenzek, line 35: I don't think this field is used at all ?

* if it's not used: please remove it, and replace `metadata.devices` by just `devices`
* if it's used: why do you have both `events` and `metadata.trace_events` ?

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,239 @@
...
+pub const xla_async_op_line_name = "Async XLA Ops";
+
+pub const host_threads_plane_name = "/host:CPU";
+pub const gpu_plane_prefix = "/device:GPU:";
+pub const tpu_plane_prefix = "/device:TPU:";
+pub const custom_plane_prefix = "/device:CUSTOM:";
+
+pub const TraceContainer = struct {
+    metadata: Trace = .{},
+    events: std.ArrayListUnmanaged(*TraceEvent) = .{},
```
gwenzek, line 31: why is this storing pointers instead of actual structs ?

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,239 @@
...
+                    if (stat_type == .step_name) event.name = stat_str;
+                    try event.args.put(allocator, xplane.getStatMetadata(xstat.metadata_id).name.getSlice(), stat_str);
+                }
+            }
+        }
+    }
+
+    fn getTraceViewerMaxEvents() !u64 {
+        const kMaxEvents = 1000000;
+        if (std.posix.getenv("TF_PROFILER_TRACE_VIEWER_MAX_EVENTS")) |max_events| {
```
gwenzek, line 171: ugh, we don't do this here. Please move this to an argument to fromXSpace, and if needed to a cli arg

pjrt/convert/trace_container.zig
```diff
@@ -0,0 +1,239 @@
...
+
+    pub fn capEvents(self: *TraceContainer, max_count: u64) void {
+        const total_count = self.events.items.len;
+        if (total_count <= max_count) {
+            // Nothing to do. Events are not known sorted after return.
+            return;
+        }
+        // sort the events according to start time.
+        // TODO: partial sort would improve performance.
+        std.mem.sort(*TraceEvent, self.events.items, {}, struct {
```
gwenzek, line 232: it feels a bit late to start dropping events. Maybe we should instead put a limit earlier on while reading events from the protobuffer. But for now it's ok to assume we can handle the input trace file. Ultimately the user control the size of the trace based on how long they run the model.

gwenzek (changes requested):

pjrt/convert/xplane_visitor.zig
```diff
@@ -0,0 +1,53 @@
+const std = @import("std");
+const xplane_proto = @import("//tsl:xplane_proto");
+const xplane_schema = @import("xplane_schema.zig");
+
+pub const XPlaneVisitor = struct {
+    plane: *const xplane_proto.XPlane,
+    event_metadata_by_id: std.AutoHashMapUnmanaged(i64, *const xplane_proto.XEventMetadata) = .{},
+    stat_metadata_by_id: std.AutoHashMapUnmanaged(i64, *const xplane_proto.XStatMetadata) = .{},
+
+    const EventTypeGetter = *const fn ([]const u8) xplane_schema.HostEventType;
```
gwenzek, line 10: those Getter declaration aren't used anymore

gwenzek (approved):
Nice, thanks.

Sorry for all the back and forth, but I think it was worth it. We went down to 700 lines from the original 1500 lines implementation.

There is a final quick fix, and can you double check that you can still convert and load the example trace I shared with you earlier ?

cryptodeal:
Correct, the example trace shared previously still converts as expected: 
[open_llama3b.trace.json](https://github.com/user-attachments/files/17804155/open_llama3b.trace.json)


And agreed, the code is so much cleaner and easier to follow after having made the suggested changes (in hindsight, I should have initially gone for a less literal direct port of the XLA implementation haha)

PR #82, opened 2024-11-14
Add custom call support, and tensor.print()

gwenzek:
* restore the integration with PJRT custom call extension
* add a generic `zmlHostBufferCallback` custom call that copies the received tensor to the host, and call a user supplied callback on it.
* add a default "print" method on `Tensor` that print the runtime value of a given tensor
* print itself has very little code, it's easy for a user to define their own print if they don't like ours.

limitations:
* only supported on Cuda atm
* pre allocates a buffer on the host to copy the content of the device buffer,
 this buffer won't be freed. There is exactly one buffer per "print" call in the IR.
* it uses cuda api, maybe I should be using `createViewOfDeviceBuffer` to make it less cuda specific

### DIFF
--- a/zml/ops.zig
+++ b/zml/ops.zig
@@ -640,12 +640,12 @@ pub fn identityCustomCall(name: [:0]const u8, input: Tensor, context: *anyopaque
 }
 
 /// At runtime the given tensor will be materialized and copied to host,
-/// so that the user can call arbitrary code on it.
-/// TODO: do the same thing but with a pjrt.Buffer, so that the user can decide if they need to copy.
+/// and the callback will be called on it.
 pub fn addHostCallback(
     callback: *const fn (HostBuffer) void,
     input: Tensor,
 ) Tensor {
+    // TODO: implement addCallback that exposes a pjrt.Buffer, so that the user can decide if they need to copy.
     if (input.getContext().target() != .cuda) return input;
 
     const len = input.byteSize();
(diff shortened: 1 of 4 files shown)


zml/ops.zig
```diff
@@ -638,3 +639,49 @@ pub fn identityCustomCall(name: [:0]const u8, input: Tensor, context: *anyopaque
...
+pub fn addHostCallback(
+    callback: *const fn (HostBuffer) void,
+    input: Tensor,
+) Tensor {
+    // TODO: implement addCallback that exposes a pjrt.Buffer, so that the user can decide if they need to copy.
+    if (input.getContext().target() != .cuda) return input;
+
+    const len = input.byteSize();
+    // Reserve memory to be able to log the runtime Buffer later during the computation.
+    // This memory is leaked, we currently have no way to tie this lifetime to the lifetime of the module being compiled.
```
Corendos, line 654: Just a note about that, this is something solved with the `api_version` 2 that we will switch to eventually

zml/tensor.zig
```diff
@@ -3629,6 +3629,27 @@ pub const Tensor = struct {
...
+
+    /// Insert code that will print the content of the given buffer at runtime.
+    /// Only for debug purpose, it has the following limitations:
+    /// * only supported on Cuda atm
+    /// * only prints the first 1024 values
+    /// * pre allocates a buffer on the host to copy the content of the device buffer,
+    /// this buffer won't be freed. You will have one buffer per "print" call in the IR.
+    /// * does device to host synchronization so it will slow down the program execution.
+    pub fn print(input: Tensor) Tensor {
+        return ops.addHostCallback(&printCallback, input);
```
Corendos, line 3641: Maybe we should log a warning or something when not running on Cuda ?

Corendos (approved):

zml/ops.zig
gwenzek: yes, can't wait for you to add support for that :-)

zml/tensor.zig
gwenzek: I mean RTFM :-)

PR #83, opened 2024-11-14
workspace: bump llvm / xla / stablehlo / pjrt plugins and restore tests

hugomano:
- workspace: new module versions added to Bazel registry:
  - deps/xla: https://github.com/openxla/xla/commit/f25e9a00b48a083ede7f2075f16ca50aebf280a0 (`20241114.0-f25e9a0`) 
This version introduces an upgraded overlay with Python repository initialization for the TSL extension. The changes were necessitated by this [XLA commit](https://github.com/openxla/xla/commit/19b4ecbe98e2c2910d6ff6532cdb64636a71e6d5#diff-df5471accc0ccfdac45fb8397fb8e1488565baf112c72487912deda919680f8aR841), which triggers a repository call. Python's hermetic version is now provided via the .bazelrc file.
  - deps/stablehlo: https://github.com/openxla/stablehlo/commit/37487a8ee0783f29eff64910702074ba52bf6c5c (`20241113.0-37487a8`)  
  - deps/llvm: https://github.com/llvm/llvm-project/commit/01d233ff403823389f8480897e41aea84ecbb3d3 (`20241113.0-01d233f`)
 This version includes an upgraded module declaration with rules_python dependency added, as LLVM prepares its overlay for Bazel v8. [Relevant LLVM changes here](https://github.com/llvm/llvm-project/commit/3903cb4695012fb85a76c83b5616f2ffe6fa10f4)
- workspace: align used versions of rules_zig in the Bazel dependency graph
- deps/pjrt-artifacts: new PJRT artifacts [`v0.2.4`](https://github.com/zml/pjrt-artifacts/releases/tag/v0.2.4): https://github.com/zml/pjrt-artifacts/commit/e1725b22289ea34403bc3bfb5aab2bc162a45978 uses [our our fork of XLA](https://github.com/zml/xla/commits/hugomano/2024_11_14)
- pjrt: expose getPluginAttributes and getPluginAttribute private APIs and stableHLOCurrentVersion public API
- pjrt: updated customCallRegistry function with the new FFI provided by the PJRT API GPU extension 
Minimal support for this feature was added, as this is not the main focus of the current PR. ([See related changes in XLA here](https://github.com/openxla/xla/commit/ef85a7bcc308313492ebc50295a8a08b4e51b8f5))
- dialects/mlir/stablehlo: expose getCurrentVersion API and fix broken stablehloVersionFromCompatibilityRequirement API
- zml/pjrtx: serialize IR to the PJRT plugin's supported HLO version. If the plugin attribute is unavailable, it falls back to the compatibility requirement StableHLO version.
- zml/tensor: tests using batching dims for gather and scatter are now re-enabled
- zml/testing: use autoPlatform to run tests instead of forcing to CPU, allowing us to tests on differents platforms


zml/testing.zig
```diff
@@ -23,7 +25,29 @@ pub fn env() zml.Platform {
...
+            }
+        } else {
+            stdx.debug.panic("ZML_TEST_TARGET set is not a valid target, {s} was provided but only {any} are availables", .{ target, available_targets });
+        }
+    } else blk: {
+        log.info("ZML_TEST_TARGET not set, using CPU", .{});
+        break :blk .cpu;
+    };
+
+    return _ctx.?.platforms.get(target).?.withCompilationOptions(_test_compile_opts);
```
gwenzek, line 50: this feel not needed. Here is the version I wanted to push:

```
return _ctx.?.autoPlatform().withCompilationOptions(_test_compile_opts);
```

gwenzek (approved):
thanks, pretty exited about the version bump.

I don't really like the env variable to chose the test runtime see my comment, but beyond that all good for me.

gwenzek (changes requested):

pjrt/pjrt.zig
```diff
@@ -140,13 +140,52 @@ pub const Api = struct {
...
+                if (api.getPluginAttribute("stablehlo_current_version")) |v| {
+                    const sliced_version = v.inner.unnamed_0.int64_array_value[0..v.inner.value_size];
+                    _ = std.fmt.bufPrint(&buf, "{d}.{d}.{d}", .{ sliced_version[0], sliced_version[1], sliced_version[2] }) catch unreachable;
+                }
+            }
+        };
+
+        state.api = self;
+        state.once.call();
+        return state.buf[0..state.buf.len];
```
gwenzek, line 159: this is not correct for when the attribute is not found, because "" isn't a valid version

pjrt/pjrt.zig
```diff
@@ -140,13 +140,52 @@ pub const Api = struct {
         };
     }
 
+    pub inline fn stableHLOCurrentVersion(self: *const Api) ?[]u8 {
```
gwenzek, line 143: this is weird

not sure I like this global variable set with a once. Maybe move this to the plugin load time ? so we can store it in the Context ?

pjrt/pjrt.zig
```diff
@@ -140,13 +140,52 @@ pub const Api = struct {
...
+        const state = struct {
+            var api: *const Api = undefined;
+            var buf: [5]u8 = undefined;
+            var once = std.once(fetch);
+
+            fn fetch() void {
+                if (api.getPluginAttribute("stablehlo_current_version")) |v| {
+                    const sliced_version = v.inner.unnamed_0.int64_array_value[0..v.inner.value_size];
+                    _ = std.fmt.bufPrint(&buf, "{d}.{d}.{d}", .{ sliced_version[0], sliced_version[1], sliced_version[2] }) catch unreachable;
```
gwenzek, line 152: this will fail as soon as there is a two number version. Maybe bump it to a bigger buffer ?

hugomano:
Close in favor of https://github.com/zml/zml/pull/98 because of this ROCm issue in XLA: https://github.com/openxla/xla/issues/19700 

All reviews below are implement in #98

PR #85, opened 2024-11-15
stablehlo: add DotAlgorithm support to stablehlo.dot_general

gwenzek:
In the API I merge the "precision" and "algorithm" attributes into a union because according to the spec,
precision must be default when algorithm is set, and precision=default means "fast" when algorithm is not set.

https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dot_general

For now it's not exposed to users because the API is a bit limited. 
In particular only a subset of algorithm are supported by openxla, and only a subset of that is supported by chips.

I was able to successfully change the accumulation type for dot on Cuda, but want to experiment more before deciding what to expose.

### DIFF
--- a/zml/tensor.zig
+++ b/zml/tensor.zig
@@ -1163,19 +1163,22 @@ pub const Tensor = struct {
             res_shape = res_shape.appendDim(rhs._shape.dim(r), rhs._shape.tag(r));
         }
 
-        const loc = lhs.getContext().mlirCtx().location(@src());
+        const mlir_ctx = lhs.getContext().mlirCtx();
+        const loc = mlir_ctx.location(@src());
         const op = dialect.stablehlo.dot_general(
-            lhs.getContext().mlirCtx(),
+            mlir_ctx,
             lhs.value(),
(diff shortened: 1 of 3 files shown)


zml/tensor.zig
```diff
@@ -1163,19 +1163,22 @@ pub const Tensor = struct {
...
             loc,
             .{
                 .lhs_batching_dimensions = lhs_batching_axes.constSlice(),
                 .rhs_batching_dimensions = rhs_batching_axes.constSlice(),
                 .lhs_contracting_dimensions = lhs_contracting_axes.constSlice(),
                 .rhs_contracting_dimensions = rhs_contracting_axes.constSlice(),
-                .precision = &.{ .DEFAULT, .DEFAULT },
+                .precision = .fast,
+                // .precision = .highest,
+                // .precision = .{ .algorithm = .{ .accumulation = .f32 } },
```
Corendos, line 1181:
```suggestion
```

Corendos (approved):

PR #92, opened 2024-11-18
zml: makeBlocks returns the inner tensor to propagate tags

gwenzek:
Now zml.ops.makeBlock returns both the mlir.Block created and also the Tensors returned by `func`.

The returned tensors should not be returned to the user,
because their `mlir.Value` must not escape the block that created them. But their shapes/tags can be safely propagated further.

This makes `zml.ops.makeBlock` a bit more footgunny, but also it's a utility function that most users shouldn't use. And it allows to use tags created inside the block and apply them to the return value. In particular tags and sharding info should now be propagated more consistently for while loops, summation, reduceWindow, ...

I've also enabled tests to run on non-cpu platform.
this is useful to figure out if a specific platform has limitation in it's pjrt support

### DIFF
--- a/zml/ops.zig
+++ b/zml/ops.zig
@@ -47,8 +47,7 @@ pub fn while_(
         @compileError("cond_fn and body_fn signatures don't match ! " ++ @typeName(@TypeOf(cond_fn)) ++ " and " ++ @typeName(@TypeOf(body_fn)));
     }
     const ctx = CompilationContext.current();
-    const cond_block, const predicate = ctx.makeBlock(CondS, &cond_fn, blkctx, inputs);
-    _ = predicate;
+    const cond_block, _ = ctx.makeBlock(CondS, &cond_fn, blkctx, inputs);
 
     const body_block, const body_res = ctx.makeBlock(BodyS, &body_fn, blkctx, inputs);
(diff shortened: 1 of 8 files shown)


zml/ops.zig
```diff
@@ -470,7 +469,8 @@ pub fn sort(
         inits[i * 2 + 1] = Tensor{ ._shape = arg_shape, ._id = undefined, ._donation = .no_buffer };
     }
     const ctx = CompilationContext.current();
-    const block = ctx.makeBlock(BodyS, &comp_fn, blkctx, inits);
+    const block, const cmp_res = ctx.makeBlock(BodyS, &comp_fn, blkctx, inits);
+    _ = cmp_res;
```
Corendos, line 473:
```suggestion
    const block, _ = ctx.makeBlock(BodyS, &comp_fn, blkctx, inits);
```

(Except if you think it's nice to convey what is the second returned value)

Corendos (approved):

zml/ops.zig
gwenzek: I don't think this work with our Zig version.

gwenzek: my bad it works, updated every where applicable

PR #96, opened 2024-11-20
examples: fixes related to recent async changes

Corendos:
Recent changes to the `async` module reverted `asyncGeneric` to `asyncc` but examples were not updated

### DIFF
--- a/examples/simple_layer/main.zig
+++ b/examples/simple_layer/main.zig
@@ -62,7 +62,7 @@ pub fn asyncMain() !void {
 
     // Start compiling. This uses the inferred shapes from the BufferStore.
     // The shape of the input tensor, we have to pass in manually.
-    var compilation = try asynk.asyncGeneric(zml.compileModel, .{ allocator, model_shapes, .forward, .{input_shape}, platform });
+    var compilation = try asynk.asyncc(zml.compileModel, .{ allocator, Layer.forward, model_shapes, .{input_shape}, platform });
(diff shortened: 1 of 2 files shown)


examples/simple_layer/main.zig
```diff
@@ -62,7 +62,7 @@ pub fn asyncMain() !void {
 
     // Start compiling. This uses the inferred shapes from the BufferStore.
     // The shape of the input tensor, we have to pass in manually.
-    var compilation = try asynk.asyncGeneric(zml.compileModel, .{ allocator, model_shapes, .forward, .{input_shape}, platform });
+    var compilation = try asynk.asyncc(zml.compileModel, .{ allocator, Layer.forward, model_shapes, .{input_shape}, platform });
```
renerocksai, line 65: We also need to update the documentation. E.g. in "Writing your first model", the `simple_layer` code is used as example.

I can deploy the docs.

gwenzek: agreed Corentin, can you do the search/replace on the full repo ?

Corendos: Sure, let me do that quickly

Corendos: Actually, we didn't update the documentation when we switched to `asyncGeneric`, so... it's already up-to-date 😂

renerocksai (approved):
YES!!!

Issue #97, opened 2024-11-20
unused tensor crashes MLIR bytecode writer


gwenzek:
```
test "unused tensor" {
    const zml = @import("zml.zig");
    const platform = zml.testing.env();

    const Local = struct {
        pub fn forward(x: Tensor) Tensor {
            const y = x.addConstant(1);
            _ = y;
            return x;
        }
    };

    const mod = try compileFn(std.testing.allocator, Local.forward, .{ Shape.init(.{ 10 }, .f32) }, platform);
    defer mod.deinit();
}
```

error:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000100952fb0 test`(anonymous namespace)::BytecodeWriter::writeUseListOrders((anonymous namespace)::EncodingEmitter&, unsigned char&, mlir::ValueRange) + 548
    frame #1: 0x0000000100956b54 test`(anonymous namespace)::BytecodeWriter::writeRegions((anonymous namespace)::EncodingEmitter&, llvm::MutableArrayRef<mlir::Region>)::'lambda'(mlir::Region&)::operator()(mlir::Region&) const + 1636
    frame #2: 0x00000001009529f8 test`(anonymous namespace)::BytecodeWriter::writeOp((anonymous namespace)::EncodingEmitter&, mlir::Operation*) + 3148
    frame #3: 0x0000000100956ba0 test`(anonymous namespace)::BytecodeWriter::writeRegions((anonymous namespace)::EncodingEmitter&, llvm::MutableArrayRef<mlir::Region>)::'lambda'(mlir::Region&)::operator()(mlir::Region&) const + 1712
    frame #4: 0x00000001009529f8 test`(anonymous namespace)::BytecodeWriter::writeOp((anonymous namespace)::EncodingEmitter&, mlir::Operation*) + 3148
    frame #5: 0x000000010094d170 test`mlir::writeBytecodeToFile(mlir::Operation*, llvm::raw_ostream&, mlir::BytecodeWriterConfig const&) + 3184
    frame #6: 0x0000000100935ed8 test`mlirOperationWriteBytecode + 208
    frame #7: 0x0000000100170bd8 test`mlir.Operation.writeBytecode__anon_30157(self=<unavailable>, writer=<unavailable>) at mlir.zig:929:37
[...]

PR #99, opened 2024-11-21
Fix crash in `for_`

gwenzek:
Contains two bug fix

* values should be pushed to their block before we open a new block, otherwise this lead to us assigning values to the wrong block.
* in `for_` make sure `first_step` is always used. This is a workaround for #97. For some reason you need two nested loops to trigger the bug. Note that this used to work, so I'm not sure what changed in how we generate MLIR or on MLIR side.


zml/ops.zig
```diff
@@ -138,7 +137,7 @@ pub fn reduce(
     var init_values: [N]mlir.Value = undefined;
     ctx.extractValues(&inits, &init_values);
 
-    const body_block, _ = ctx.makeBlock(BodyS, &body_fn, {}, .{ inits, inits });
+    const body_block, _ = ctx.makeBlock(.hermetic, BodyS, &body_fn, {}, .{ inits, inits });
```
Corendos, line 140: Just curious, `reduce` operations are not allowed to access outer scope ?

Corendos (approved):

zml/ops.zig
gwenzek: I thought so, but I tested generating some IR that was using a constant from above and it worked. But since `zml.ops.reduce` doesn't take a context, it's not possible in ZML.
Maybe we should add a `reduceWithCtx` ?

gwenzek: at least stablehlo-opt doesn't complain, but I haven't tried actually running anything like that. Maybe leave it like that until we have a usecase ?

Corendos: One usecase I see would be a `scan` like function that could prove useful for flash attention
But yeah, in the meantime we can keep it like this

PR #100, opened 2024-11-22
async: rework async runtime to use coroutines and cross thread signaling

steeve:
This mostly reworks the runtime to enable proper cross thread signaling for `callBlocking`.

API changes include:
- `async_ -> asyncc`
- `await_ -> awaitt`
- proper type inference for non generic arguments even for generic functions
- bump `libxev`
- use `epoll` on linux by default (can be comptime overridden through `xev_options.linux_backend` in the root module)

Mostly, also, it's pretty fast.


gwenzek:
LGTM, just confused about the PR description. For me `async_` is already gone on master ?

async/zigcoro.zig
```diff
@@ -1,123 +1,203 @@
...
+        const self: *FrameAllocator = @ptrCast(@alignCast(ctx));
+        const stack = self.pool.create() catch return null;
+        return @ptrCast(stack);
     }
 
-    pub fn wait(self: *Notification) !void {
-        try self.inner.wait();
+    fn resize(ctx: *anyopaque, buf: []u8, buf_align: u8, new_len: usize, ret_addr: usize) bool {
+        _ = ctx;
+        _ = buf;
```
gwenzek, line 164: you should allow this:  `return new_len <=Item.len;`

async/zigcoro.zig
```diff
@@ -1,123 +1,203 @@
...
+                .resize = resize,
+                .free = free,
+            },
         };
     }
 
-    pub fn notify(self: *Notification) !void {
-        try self.inner.notif.notify();
+    fn alloc(ctx: *anyopaque, len: usize, ptr_align: u8, ret_addr: usize) ?[*]u8 {
+        _ = len;
```
gwenzek, line 154: `std.debug.assert(len <= Item.len);`

async/zigcoro.zig
```diff
@@ -1,123 +1,203 @@
...
+                .waiting = &waiter,
+            };
+            if (self.waiter.cmpxchgStrong(&State.unset_state, &new_state, .monotonic, .monotonic) == null) {
+                libcoro.xsuspend();
+            }
+        }
+    };
+};
+
+pub const FrameAllocator = struct {
```
gwenzek, line 130: who is using this allocator ? I didn't see any async code using it directly

steeve: yeah it's not really used, I kept it because it could prove useful

PR #102, opened 2024-11-26
runtimes/neuron: add support for AWS neuron / trainium / inferentia 2

steeve:
This PR adds support for AWS Neuron / Trainium 1 / Inferentia 2.

It can be enabled via the `--@zml//runtimes:neuron=true` flag.

Everything should work as expected.


runtimes/BUILD.bazel
```diff
@@ -25,6 +28,39 @@ RUNTIMES = {
     for runtime in RUNTIMES.keys()
 ]
 
+write_file(
```
gwenzek, line 31: what's that ?

runtimes/BUILD.bazel
```diff
@@ -25,6 +28,39 @@ RUNTIMES = {
...
+    name = "zml.txt",
+    out = "zml.txt",
+    content = ["ZML loves you <3"],
+)
+
+tar(
```
gwenzek, line 37: and this ?

runtimes/cpu/BUILD.bazel
```diff
@@ -25,3 +25,9 @@ zig_library(
         "//conditions:default": [":empty"],
     }),
 )
+
+filegroup(
```
gwenzek, line 29: why ?

gwenzek (approved):
Noice !

runtimes/BUILD.bazel
steeve: needed because else `oci_image` complains that `layers` is empty

PR #107, opened 2024-11-27
workspace: log message before return NotFound 

hugomano:
This allow us to find which NotFound is raised

-- 
~~require #106 to be merged before~~

### DIFF
--- a/zml/context.zig
+++ b/zml/context.zig
@@ -98,7 +98,11 @@ pub const Context = struct {
                 num_platforms += 1;
             }
         }
-        if (num_platforms == 0) return error.NotFound;
+        if (num_platforms == 0) {
+            log.err("Zero platform available", .{});
+            return error.NotFound;
+        }
+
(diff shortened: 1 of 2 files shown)


zml/context.zig
```diff
@@ -98,7 +98,11 @@ pub const Context = struct {
                 num_platforms += 1;
             }
         }
-        if (num_platforms == 0) return error.NotFound;
+        if (num_platforms == 0) {
+            log.err("Zero platform available", .{});
+            return error.NotFound;
```
gwenzek, line 103: arguably NotFound is not the best name. Maybe "NoPlatformAvailable" ?

gwenzek (approved):

zml/context.zig
hugomano: sure I change that, sometimes I forgot the majesty of custom errors and error union in Zig

PR #111, opened 2024-11-28
runtimes/neuron: use the cloudflare debian snapshot endpoints

steeve:
The regular endpoints can remove some packages.


gwenzek (approved):

gwenzek:
If you're looking at Neuron, could you also fix https://github.com/zml/zml/security/dependabot/1 ?

PR #116, opened 2024-12-02
zml: emit `func.call` operations

gwenzek:
Restore the machinery to emit individual functions and calls to them.

this should unblock composites, and generates cleaner MLIR in general

* functions are cached based on the input shapes including tags.
* donations/tags are propagated through function calls

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -481,8 +481,10 @@ pub const CompilationContext = struct {
         const Local = struct {
             bias: Tensor,
 
-            pub fn forward(self: @This(), x: Tensor) Tensor {
-                return zml.ops.call(self, .inner, .{x});
+            pub fn forward(self: @This(), x: Tensor, y: Tensor) [2]Tensor {
+                const x1 = zml.ops.call(self, .inner, .{x});
+                const x2 = zml.ops.call(self, .inner, .{x1});
+                return .{ x1.reuseBuffer(y), x2 };
             }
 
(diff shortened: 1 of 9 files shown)


.github/workflows/commit.yaml
Corendos, line 1: This should probably be put in another commit/PR to have a proper history

zml/module.zig
```diff
@@ -49,7 +49,7 @@ const Block = union(BlockKind) {
             .op_result => |parent_op| self.appendOperationRecursive(parent_op),
             .block_argument => |arg| {
                 // Hermetic blocks are not allowed to use arguments from other blocks.
-                std.debug.assert(self == .open or self.block().eql(arg.block()));
+                stdx.debug.assert(self == .open or self.block().eql(arg.block()), "Can't add {}  from {?x} block to {?x} block", .{ arg, arg.block()._inner.ptr, self.block()._inner.ptr });
```
Corendos, line 52:
```suggestion
                stdx.debug.assert(self == .open or self.block().eql(arg.block()), "Can't add {} from {?x} block to {?x} block", .{ arg, arg.block()._inner.ptr, self.block()._inner.ptr });
```

Corendos (approved):

.github/workflows/commit.yaml
gwenzek: done: https://github.com/zml/zml/pull/121

gwenzek: done https://github.com/zml/zml/pull/121

PR #117, opened 2024-12-02
Llama 3.2 (text only models) support

AdjectiveAllison:
llama 3.2 1B and 3B do not have lm_head. 

I've added transposing over the embed_tokens as a replacement and set lm_head as optional for backward compatibility with the other models. 

I also added both repositories+executions to bazel and updated the README.

### DIFF
--- a/examples/llama/llama.zig
+++ b/examples/llama/llama.zig
@@ -24,7 +24,7 @@ pub const LlamaOptions = struct {
 /// Llama architecture, using huggingface transformers naming.
 /// Dimensions of activations: {.b, .s, .d}
 pub const LlamaLM = struct {
-    lm_head: zml.nn.Linear,
+    lm_head: ?zml.nn.Linear = null,
     model: Llama,
 
     // Options controlling generation
@@ -55,7 +55,9 @@ pub const LlamaLM = struct {
         // TODO(Corentin): Fix lm_head sharding when top-k sampling is enabled.
         // It currently crashes/compilation fails
         if (options.gen_opts.topk == 1) {
-            self.lm_head.weight = self.lm_head.weight.withSharding(.{0});
+            if (self.lm_head) |lm_head| {
+                self.lm_head.?.weight = lm_head.weight.withSharding(.{0});
+            }
         }
     }
 
@@ -76,12 +78,12 @@ pub const LlamaLM = struct {
 
         var tokens = tokens_.withPartialTags(.{.s});
         const out, const updated_kv_cache = zml.call(self.model, .forward, .{ tokens, if (kv_cache == null) null else token_index, kv_cache });
-        tokens, const new_rng = updateTokens(self.lm_head, tokens, token_index, out, rng, self.gen_opts);
+        tokens, const new_rng = self.updateTokens(tokens, token_index, out, rng, self.gen_opts);
         return .{ tokens, increment(0, token_index), updated_kv_cache, new_rng };
     }
 
     pub fn updateTokens(
-        lm_head: zml.nn.Linear,
+        self: LlamaLM,
         tokens_: Tensor,
         token_index: Tensor,
         out_: Tensor,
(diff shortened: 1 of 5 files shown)


AdjectiveAllison:
That last commit may be a lil bit of scope creep, but I noticed llama 3.2 but often hit eos tokens (the same ones 3.1 has) and print out the remaining of the `token_buffer` (filled with 0) and subsequently calculate speed wrong. I tested a whole variety of use cases and the eos handling seems to work across them all (as does the output printing and token count index regardless of whether eos was hit or not).

gwenzek:
Thanks @AdjectiveAllison looks great.
For the early eos stop can you try reusing the eos token from the tokenizer ?
I don't think it's a good idea to hardcode tokens like that, but also I understand it's a bit annoying. 
FWITW serious Llama server typically have a user flag to chose when to stop.

AdjectiveAllison:
> Thanks @AdjectiveAllison looks great. For the early eos stop can you try reusing the eos token from the tokenizer ? I don't think it's a good idea to hardcode tokens like that, but also I understand it's a bit annoying. FWITW serious Llama server typically have a user flag to chose when to stop.

Yeah I totally agree. Would your suggested route to take be to use `tokenizer.special_tokens.eos`? This returns back `128001` as expected, but the issue I am having is that one of the other eos tokens (according to the [config.json](https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct/blob/main/config.json)) is the one I'm getting the eos on (128008), and thus the code wouldn't function. This same triplet of eos tokens (128001, 128008, 128009) are in the [llama 3.1 8B config.json](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/blob/main/config.json) as well. 

It looks like both repositories also have a [special_tokens_maps.json](https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct/blob/main/special_tokens_map.json) file in them that highlight 128008 as the desired eos token, but the tokenizer implementation in zml simply looks to string match and best-guess on the correct id ([seen here](https://github.com/zml/zml/blob/a6ea0bd90568ce20537078ee61bdf978d7d10c13/zml/tokenizer.zig#L999)). 

Starting to look like I scope creeped more than intended. These are the directions that I can go from here:

1. Leave hardwired ids currently ([It was hardwired before my change as well](https://github.com/zml/zml/blob/a6ea0bd90568ce20537078ee61bdf978d7d10c13/examples/llama/main.zig#L101), just incorrectly for llama 3.1/3.2 and not used for speed calculation and generation stopping)
2. Come up with solution for handling eos tokens better
  a. Quickly patch the string matching eos tokenizer adding code to look for <|eot_id|> first so 128008 takes precedence)
  b. Make `eos` tokens be an array, and allow it to be passed in alongside the config in bazel?
  c. Other, more fully thought out solution.
3. Remove eos commit and leave this PR as-is with llama 3.2 support. Potentially can work on 2 in another issue/pr. 

That's a good call out on the way serious llama servers handle this. I'll have to go give a peak at what they're doing here.

AdjectiveAllison:
https://github.com/zml/zml/issues/119

I threw this issue up to discuss more properly what a solution looks like. Let me know if you want for this current llama 3.2 PR (retract eos changes or keep as-is). Thanks for pointing it out @gwenzek

examples/llama/llama.zig
```diff
@@ -92,7 +94,11 @@ pub const LlamaLM = struct {
         const out = out_.withPartialTags(.{ .s, .d });
 
         const next_token_pred = out.gatherValues(.s, token_index, .{});
-        var logits = zml.call(lm_head, .forward, .{next_token_pred});
+        var logits = if (self.lm_head) |lm_head|
+            zml.call(lm_head, .forward, .{next_token_pred})
+        else
+            next_token_pred.matmul(self.model.embed_tokens.weight.transpose(.{ -1, -2 })).withTags(.{.d});
```
gwenzek, line 100: can you try using dot and tags ? it should look something like

```
self.model.embed_tokens.weight.withTag(.{.voc, .d}).dot(next_token_pred, .d);
```

examples/llama/main.zig
```diff
@@ -86,26 +87,44 @@ pub fn generateText(
             decode_progress += output_freq;
             std.debug.print("{s}", .{output.items[n..]});
             tracer.frameEnd(frame_id, try std.fmt.bufPrintZ(tracer_buffer, "Decoded token {}/{} : {s}", .{ i + 1, output_tokens_len, output.items[n..] }));
+
+            // These are almost certainly not the most optimal way to detect an EOS token.
+            // Do tinyLlama or openLlama use different tokenizers? Can we get these eos values automatically?
+            if (std.mem.indexOfScalar(i32, token_buffer[decode_progress..], 128001)) |_| {
```
gwenzek, line 93: ok let's keep this, since the demo is mostly made for llama model family it's fine to hardocode llama eos tokens.

can you use though `std.mem.indexOfAny` to match all 3 eos tokens in one go ?

gwenzek (approved):
Thanks a lot ! just tested on my laptop and it works great :-)

PR #122, opened 2024-12-06
async: add intrusive queue

gwenzek:
all code contributed by @steeve

* add intrusive queue
* change the constructor of Channel with default AsyncThread executor


renerocksai (approved):
LGTM

PR #115, opened 2024-11-29
zml: cleanup module.zig

gwenzek:
* compileInternal now works with function not with "module"
* BaseExe now allocates all the memory needed to call `pjrtExecute`
* `Exe` is the type variant of `BaseExe` and enforce the right Zig arg are passed to `call`. `exe.prepare(model)` don't need to allocate anymore, the allocation happened during compilation.
* public entry points like `compile`, `compileModel`, `compileFn` are now at the top of the file
* the caching logic has been mostly inlined inside `compileInternal`
* `compile`, `compileFn`, `Exe`, `BaseExe` which are the user visible API have been moved to `exe.zig`
* `CompilationContext` and all the compilation logic stays in `module.zig`

PR #123, opened 2024-12-06
update instructions

gwenzek:
following #115, `prepare` doesn't alloc anymore, `ExeWithWeights` is `ModuleExe`


renerocksai (approved):
Thanks for updating!

PR #124, opened 2024-12-06
Fix test layer

gwenzek:
* testLayer was silently broken by #115 : test + fix so we detect this in the future. 
* mapAlloc had a regression in #115 wrt struct with 0 size fields: test + fix

* add Python `venv` to gitignore


hugomano (approved):
Thanks!

Issue #125, opened 2024-12-07
CUDA GPU memory allocation percentage does not dynamically scale


AdjectiveAllison:
## Description
The default cuda GPU memory allocation is set to 75% and doesn't dynamically scale even when memory is available, causing OOM errors that can be resolved by manually increasing the memory fraction to 0.95.

## Reproduction
Using Llama-3.1-8B-Instruct with a sequence length of 15000 (with 48GB of vram):

Fails with OOM:
```bash
bazel run -c opt //llama:Llama-3.1-8B-Instruct --@zml//runtimes:cuda=true --@zml//runtimes:cpu=false -- --seq-len=15000 
```

Same command works successfully when memory fraction is set to 0.95 (@gwenzek helped me set this manually in a discord thread)
```bash
bazel run -c opt //llama:Llama-3.1-8B-Instruct --@zml//runtimes:cuda=true --@zml//runtimes:cpu=false -- --seq-len=15000
```

Can simply raise seq-len on any cuda device until it OOM, and then raise memory fraction and see it works.

## Environment
```
NVIDIA-SMI 565.57.01    Driver Version: 565.57.01    CUDA Version: 12.7
GPU: NVIDIA L40S (49140MiB total memory)
```

## Observed Behavior
1. GPU memory usage is capped at exactly 75% (36843MB/49140MB) 
2. OOM error occurs trying to allocate ~28.2GB additional memory:
```
...
2024-12-07 00:50:29.492423: I external/xla/xla/tsl/framework/bfc_allocator.cc:1117] 2 Chunks of size 983040000 totalling 1.83GiB
2024-12-07 00:50:29.492425: I external/xla/xla/tsl/framework/bfc_allocator.cc:1117] 2 Chunks of size 1050673152 totalling 1.96GiB
2024-12-07 00:50:29.492428: I external/xla/xla/tsl/framework/bfc_allocator.cc:1121] Sum Total of in-use chunks: 16.79GiB
2024-12-07 00:50:29.492431: I external/xla/xla/tsl/framework/bfc_allocator.cc:1123] Total bytes in pool: 38171934720 memory_limit_: 38171934720 available bytes: 0 curr_region_allocation_bytes_: 76343869440
2024-12-07 00:50:29.492437: I external/xla/xla/tsl/framework/bfc_allocator.cc:1128] Stats:
Limit:                     38171934720
InUse:                     18026663936
MaxInUse:                  18026663936
NumAllocs:                         299
MaxAllocSize:               1050673152
[...]

PR #129, opened 2024-12-10
pjrt/zml: expose client creation flags

gwenzek:
The Pjrt api allows to passe string flags to the plugin on client creation.
Since we don't like working with strings I've converted the strings supported by the cuda plugin into a struct.
For now I've focused on the strings corresponding to the allocator settings following #125 let me know if you think 
that we should expose more from:

https://github.com/openxla/xla/blob/3d31c48c719d331d432132b3e0c2c5ce52650675/xla/pjrt/c/pjrt_c_api_gpu_internal.cc#L76-L86

AFAICT the cpu plugin doesn't read flags at all.

Will need some investigation for TPU and neuron platforms

**Breaking change**: `context.autoPlatform()` -> `context.autoPlatform(.{})`

### DIFF
--- a/zml/platform.zig
+++ b/zml/platform.zig
@@ -30,8 +30,8 @@ pub const Platform = struct {
 
     pub const MAX_NUM_DEVICES: u8 = 32;
 
-    pub fn init(target: Target, api: *const pjrt.Api) !Platform {
-        const pjrt_client = try pjrt.Client.init(api, &.{});
+    pub fn init(target: Target, api: *const pjrt.Api, options: pjrt.Client.CreateOptions) !Platform {
+        const pjrt_client = try pjrt.Client.init(api, options);
         const true_num_devices = pjrt_client.getAddressableDevices(api).len;
         if (true_num_devices > MAX_NUM_DEVICES) {
             log.warn("platform {} got {} devices, but ZML only support up to {} devices. Some devices won't be used.", .{ target, true_num_devices, MAX_NUM_DEVICES });
--- a/zml/platform.zig
+++ b/zml/platform.zig
@@ -29,10 +29,11 @@ pub const Platform = struct {
     compilation_options: CompilationOptions = .{},
 
     pub const MAX_NUM_DEVICES: u8 = 32;
-    pub const CreateOptions = pjrt.Client.CreateOptions;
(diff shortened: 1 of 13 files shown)


gwenzek:
I'm not sure how I feel about the change from `void` to `NoOpt` it's a bit of no-op from user persepective, but it's really to come around limitations in `json.parse` for struct with `void` field.

AdjectiveAllison:
>  let me know if you think that we should expose more from:
> https://github.com/openxla/xla/blob/3d31c48c719d331d432132b3e0c2c5ce52650675/xla/pjrt/c/pjrt_c_api_gpu_internal.cc#L76-L86

Idk if this is addressed to me but here is my 2 cents: 

`num_nodes` could be good! I found myself on a multi gpu system that had interesting performance characteristics with different models and I was wondering if some of the smaller models would run slower/faster if limited to a single gpu.

zml/platform.zig
```diff
@@ -84,3 +86,78 @@ pub const Platform = struct {
...
+    // NoOpt instead of void to allow it to work with json. Yes, it's a bit annoying.
+    pub const NoOpt = struct {};
+
+    pub const Cuda = struct {
+        // Use the same default than XLA.
+        allocator: Allocator = .{ .bfc = .{} },
+        // TODO support all of https://github.com/openxla/xla/blob/3d31c48c719d331d432132b3e0c2c5ce52650675/xla/pjrt/c/pjrt_c_api_gpu_internal.cc#L76-L86
+        visible_devices: []const u16 = &.{},
+        // node_id
+        // num_nodes
```
renerocksai, line 111: @AdjectiveAllison suggested adding at least this ^^^ `num_nodes`

renerocksai (changes requested):

renerocksai (approved):

zml/platform.zig
gwenzek: I agree with the feeling,
I've started https://github.com/zml/zml/pull/132 but it's not working as expected

in the interest of moving forward, I'll merge this PR and follow up later

gwenzek:
Note: I've started https://github.com/zml/zml/pull/132 to add devices flags but it's not working as expected

in the interest of moving forward, I'll merge this PR and follow up later

PR #130, opened 2024-12-10
zml: add an arena to the CompilationContext

gwenzek:
* add an arena to the CompilationContext

There was already an allocator in the CompilationContext, this PR makes it into an arena and expose the arena allocator.
This allows ops that need to allocate to do so without abusing `std.BoundedArray` with arbitrary bounds, and in a safe way.
For now little ops use it, which I believe should stay the case, because most of the time you want to know how many tensor you're working with.
It'll also simplify the life of Zig new-comers

this include some **breaking changes** to `chunkAllowTrailing` and `split` that are allocating

* upgrade a few `axis_: i64` into `axis_: anytype` to handle tags

* add a few TODO for the next version of the Tensor API

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -145,7 +145,7 @@ pub const CompilationContext = struct {
     }
 
     pub fn deinit(self: *CompilationContext) void {
-        // self._fn_cache.deinit(self.allocator());
+        // No need to deinit self._fn_cache cause it uses our arena
         self._mlir_ctx.deinit();
         self._mlir_registry.deinit();
         self._arena.deinit();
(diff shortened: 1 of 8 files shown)


zml/module.zig
```diff
@@ -135,17 +138,21 @@ pub const CompilationContext = struct {
...
-            ._allocator = allocator,
+            ._fn_cache = .{},
+            ._arena = arena,
             ._tracer = Tracer.init("ai.zml.compilation"),
         };
     }
 
     pub fn deinit(self: *CompilationContext) void {
-        self._fn_cache.deinit();
+        // self._fn_cache.deinit(self.allocator());
```
Corendos, line 148:
```suggestion
```

gwenzek: replace by a more explicit comment

PR #135, opened 2024-12-12
zml: Reorganize the compilation output folders

gwenzek:
* create one folder per target, then one sub-folder per mlir module.
* pass this folder to `xla_dump_to` so that all the xla files arrive there.
* explicitly save the mlir and the serialized executable there

I'm also exposing James profile conversion tool into ZML as `profiler.dumpDataAsJson(file)`

### DIFF
--- a/pjrt/pjrt.zig
+++ b/pjrt/pjrt.zig
@@ -359,7 +359,7 @@ pub const Client = opaque {
             }
         }
         log.warn("No profiler found for platform: {}", .{self});
-        return Profiler.init(null, options);
+        return Profiler.init(null, null);
     }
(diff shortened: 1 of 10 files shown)


pjrt/pjrt.zig
```diff
@@ -359,7 +359,7 @@ pub const Client = opaque {
             }
         }
         log.warn("No profiler found for platform: {}", .{self});
-        return Profiler.init(null, options);
+        return Profiler.init(null, null);
```
Corendos, line 362: Why not using options ?

Corendos (approved):

PR #142, opened 2025-01-06
zml: rewrite simple transpose

gwenzek:
Rewrite trivial transpose into reshapes.
Transpose over a 1-d axis are effectively no-op.
Normally the backend can detect this, but we observe that some backend can get confused. So avoiding those extra transpose can help.

I also changed the default profiler size to 1_000_000 events because above that the perfetto ui can't load the generated file anyway


renerocksai (approved):
LGTM && works on my machine

PR #146, opened 2025-01-07
pjrt: Profiler fixes

gwenzek:
* allow i64 resource id
* reserve memory when creating array lists


cryptodeal (approved):
Looks good to me; good call to reserve memory for ArrayLists ^

PR #147, opened 2025-01-08
zml: HostBuffer.prettyPrint()

gwenzek:
Add pretty printing of HostBuffer.

This will be leverage by the debug helper `x.print()`
It can also be used like this: `std.log.info("my buffer: {}", .{host_buffer.pretty()})`

### DIFF
--- a/zml/testing.zig
+++ b/zml/testing.zig
@@ -53,6 +53,7 @@ pub fn expectClose(left_: anytype, right_: anytype, tolerance: f32) !void {
         if (should_free_left) left.deinit(allocator);
         if (should_free_right) right.deinit(allocator);
     }
+    errdefer log.err("\n--> Left: {}\n--> Right: {}", .{ left.pretty(), right.pretty() });
 
     if (!std.mem.eql(i64, left.shape().dims(), right.shape().dims())) {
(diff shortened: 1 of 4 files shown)


zml/testing.zig
```diff
@@ -13,19 +13,17 @@ var _platform: ?zml.Platform = null;
...
     if (!builtin.is_test) @compileError("Cannot use zml.testing.env outside of a test block");
     if (_platform == null) {
-        _test_compile_opts = .{
-            .sharding_enabled = true,
-        };
-
         var ctx = zml.Context.init() catch unreachable;
-        _platform = ctx.autoPlatform(.{}).withCompilationOptions(_test_compile_opts);
+        _platform = ctx.autoPlatform(.{}).withCompilationOptions(.{
+            .xla_dump_to = "/tmp/zml/tests/",
```
Corendos, line 18: We might want to consider an option to opt-out from module caching when we use the `xla_dump_to` because we don't want that for tests

Corendos (approved):

zml/testing.zig
gwenzek: yes on providing the opt-out, but for tests it's pretty good to have the full test suite running in seconds when you iterate on a small part of the library

PR #153, opened 2025-01-15
async,runtimes: bump rocm, fixe cuda and rocm sandboxes, bump libxev

steeve:
This PR is a collection of fixes related to CUDA and ROCm sandbox discovery.

In the process, update the epoll libxev patch which caused 100% CPU on linux.

While fixing CUDA add `xla_gpu_enable_latency_hiding_scheduler` as per NVIDIA's recommendations.

Finally, bump cuDNN to 9.6.0 while we're here.


runtimes/cuda/cuda.zig
```diff
@@ -12,6 +17,23 @@ fn hasNvidiaDevice() bool {
     return true;
 }
 
+fn setupXlaGpuCudaDirFlag() !void {
```
gwenzek, line 20: I find code easier to read when this kind of internal helper function is located at the bottom of the file,
or at least after their call site

runtimes/rocm/rocm.zig
```diff
@@ -14,6 +19,44 @@ fn hasRocmDevices() bool {
     return true;
 }
 
+fn setupRocmEnv() !void {
```
gwenzek, line 22: same here, it's better to keep `load` which is the main entry point at the top of the file

gwenzek (approved):

PR #133, opened 2024-12-11
Llama tokenizer has issues with emojis

renerocksai:
Using openwebui, the following prompt template is used:

```
Create a concise, 3-5 word title with an emoji as a title for the chat history, in the given language. Suitable Emojis for the summary can be used to enhance understanding but avoid quotation marks or special formatting. RESPOND ONLY WITH THE TITLE TEXT.

Examples of titles:
📉 Stock Market Trends
🍪 Perfect Chocolate Chip Recipe
Evolution of Music Streaming
Remote Work Productivity Tips
Artificial Intelligence in Healthcare
🎮 Video Game Development Insights

<chat_history>
{{MESSAGES:END:2}}
</chat_history>
```

This causes the llama tokenizer to crash.

My fix mitigates the issue but I am not sure it is correct.

PR #157, opened 2025-01-20
Llama3.2: fix tokenizer

### LINKED ISSUE #134: Llama tokenizer has issues with emojis
See #133

gwenzek:
Fix #134 , superseed #133

Basically the llama3 tokenizer.json has the following config: 

```
  "model": {
    "type": "BPE",
    "dropout": null,
    "unk_token": null,
    "continuing_subword_prefix": null,
    "end_of_word_suffix": null,
    "fuse_unk": false,
    "byte_fallback": false,
    ...
}
```

this it technically invalid because it says the model has no byte fallback and no unknown token either.
This means that unknown emoji can't be converted to bytes nor to the special unknown token.
This ended up in tokenizer.zig using the value of MAX_INT for the UNK token then panicking for out of bound.

So I've added some logic to recognize that in the case of GPT2 encoded vocab (like llama3) that actually contains individual bytes, we should use the "byte_fallback" algorithm even if disabled.

I've also added a Llama3 "well known" normalizer because HF tokenizers doesn't add a dummy prefix for llama3 tokenizer.

Technically this is still not a correct implementation of llama3 tokenizer because we don't do the regex based splitting they do. But even for quite long snippets I tested it now produces exact same tokens.

### DIFF
--- a/zml/tokenizer.zig
+++ b/zml/tokenizer.zig
@@ -164,7 +164,8 @@ pub const Tokenizer = struct {
             // Step by step visualization of the progress.
             if (options.debug) {
                 var _debug_buf: [256]u8 = undefined;
-                var debug_progress: std.ArrayList(u8) = .{ .items = _debug_buf[0..0], .capacity = _debug_buf.len, .allocator = undefined };
+                var _debug_alloc = std.heap.FixedBufferAllocator.init(&_debug_buf);
+                var debug_progress = std.ArrayList(u8).init(_debug_alloc.allocator());
                 self.decodeWithOpts(&debug_progress, tok_buff[0..num_tokens], .{ .sep = "|" }) catch {};
                 log.debug("tokens: {d} -> {s}", .{ tok_buff[0..num_tokens], debug_progress.items });
             }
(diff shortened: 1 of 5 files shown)


zml/tokenizer.zig
```diff
@@ -920,15 +929,16 @@ pub fn fromHfJson(allocator: std.mem.Allocator, tokenizer_path: []const u8) !Tok
         else => return error.InvalidFormat,
     };
 
+    const model = objectGet(main_object, .object, "model") orelse return error.InvalidFormat;
     // TODO: remove all panics
```
renerocksai, line 933: is this comment still valid, I didn't see panics in the diff context?

renerocksai (approved):
Nice! Sad we have to account for misleading tokenizer.json, though.

zml/tokenizer.zig
gwenzek: so I had removed most of the .? but had forgotten some.
I remove the last ones and added a more precise TODO wrt EOS/BOS tokens

PR #158, opened 2025-01-21
Async Buffers & HF Tokenizers

renerocksai:
# async buffers, HF tokenizers, Llama improvements

This PR introduces breaking changes to the ZML APIs.

1. Buffers and execute are now asynchronous
2. Added HuggingFace tokenizer bindings, removed old tokenizers
3. Improved Llama example

## 1. zml: make buffers and execute asynchronous

This change makes buffer and execution completely asynchronous. This enables easy pipelining of multiple operations. It doesn't map well to `async` because the PJRT API is already asynchronous, so might as well use it as is.

As a consequence, you have to call `.awaitt()` on Buffers, like in this snippet taken from the `llama` example:

```zig
// Here we will copy the generated token from device
var generated_token_buffer = [_]u32{0};

// ...

// extract the generated token from the buffer, async
var fut_tok_buf = try current_token.toHost(std.mem.sliceAsBytes(&generated_token_buffer));
//  ^^^ the HostBuffer created above is now ASYNC!

// wait until the token is actually in the host buffer
_ = try fut_tok_buf.awaitt();

// after awaiting it, we can access the updated buffer
const generated_token = generated_token_buffer[0];  // <- the u32 token
```

## 2. zml/tokenizer: add HuggingFace tokenizers binding, remove old tokenizers

`zml.Tokenizer` now uses the new HuggingFace tokenizer bindings when loading a `tokenizer.json`.

To move fast, we dropped the `tinyllama`, `gguf`, and `sentencepiece` tokenizers.

## 3. examples/llama

Changes to the llama example:

- using async buffers like described above
- config.json support -> less cmdline flags
- switched to [zig-clap](https://github.com/Hejsil/zig-clap) for cmdline args parsing
- using HuggingFace tokenizer
- re-worked and simplified the code
    - single token prediction
- removed esoteric models from BUILD.bazel
- default container now uses Llama-3.2-1B-Instruct


zml/buffer.zig
```diff
@@ -27,12 +27,29 @@ const log = std.log.scoped(.zml);
 /// * loading weights from disk directly to the `device zml.aio.loadBuffers`
 /// * can be created by calling `HostBuffer.toDevice(platform)`.
 pub const Buffer = struct {
+    pub const Shard = struct {
+        api: *const pjrt.Api,
```
gwenzek, line 31: why do you need a copy of the api here ? can we have a buffer made of shards across different runtimes ?

zml/buffer.zig
```diff
@@ -27,12 +27,29 @@ const log = std.log.scoped(.zml);
...
+        buffer: *pjrt.Buffer,
+        ready_event: ?*pjrt.Event = null,
+        ready: bool = false,
+
+        pub fn awaitt(self: *Shard) !void {
+            if (self.ready) {
+                return;
+            }
+            if (self.ready_event orelse self.buffer.getReadyEvent(self.api)) |ev| {
```
gwenzek, line 40: what is `getReadyEvent` doing ? when can we create a Shard without an event ?

zml/exe.zig
```diff
@@ -210,11 +210,11 @@ pub const BaseExe = struct {
             .non_donatable_input_indices = &.{},
         }) catch unreachable;
 
-        for (events[0..sharding.num_partitions]) |e| {
-            if (e) |ev| {
-                ev.await_(self.platform.pjrt_api) catch unreachable;
-            }
-        }
+        // for (events[0..sharding.num_partitions]) |e| {
```
gwenzek, line 213: please remove this, maybe leave a comment explaining we don't synchronize execution

zml/exe.zig
```diff
@@ -316,7 +316,7 @@ fn fillBuffers(v: anytype, buffers: []const [*]*pjrt.Buffer, start: u32) u32 {
             const model_sharding = ctx.buffers.len;
             stdx.debug.assert(buffer._shards.len == model_sharding, "Can't feed a {}-sharded tensor into a {}-sharded model", .{ buffer._shards.len, ctx.buffers.len });
             for (buffer._shards.constSlice(), 0..) |shard, d| {
-                ctx.buffers[d][ctx.index] = shard;
+                ctx.buffers[d][ctx.index] = shard.buffer;
```
gwenzek, line 319: shouldn't we also set the event here ?

zml/hostbuffer.zig
```diff
@@ -251,6 +256,20 @@ pub const HostBuffer = struct {
         };
     }
 
+    pub fn awaitt(self: *HostBuffer) !*HostBuffer {
+        if (self._ready) {
```
gwenzek, line 260: I'm not sure how I feel about this, it's easy to forget to call await.
Can we call it ourselves in a few places ? like format, or items ?

zml/hostbuffer.zig
```diff
@@ -251,6 +256,20 @@ pub const HostBuffer = struct {
...
+            return self;
+        }
+        if (self._event) |ev| {
+            if (self._api) |api| {
+                try ev.awaitt(api);
```
gwenzek, line 265: should you modify the HostBuffer here to mark it as ready ?

zml/hostbuffer.zig
```diff
@@ -251,6 +256,20 @@ pub const HostBuffer = struct {
         };
     }
 
+    pub fn awaitt(self: *HostBuffer) !*HostBuffer {
+        if (self._ready) {
+            return self;
+        }
+        if (self._event) |ev| {
```
gwenzek, line 263: this logic would be simpler if we used the "make invalid state unrepresentable" strategy:

`ready: union { ready, ev: struct { *Event, *Api } };`

zml/pjrtx.zig
```diff
@@ -145,7 +196,8 @@ pub const Event = opaque {
         return self.inner().getEventError(api);
     }
 
-    pub fn await_(self: *Event, api: *const Api) !void {
+    pub const await_ = awaitt;
```
gwenzek, line 199: can we chose a syntax and stick to it, I feel we flip flop every two month on that

zml/tensor.zig
```diff
@@ -2027,7 +2027,7 @@ pub const Tensor = struct {
 
     /// Appends a 1-dim axis, with the given tag.
     pub fn appendAxes(self: Tensor, t: anytype) Tensor {
-        stdx.debug.assert(self.rank() < Tensor.MAX_RANK - t.len, "appendAxis expects tensor rank to be small enough in order to extend it, got {} and {} (max is {})", .{ self.rank(), t.len, Tensor.MAX_RANK });
+        // stdx.debug.assert(self.rank() < Tensor.MAX_RANK - t.len, "appendAxis expects tensor rank to be small enough in order to extend it, got {} and {} (max is {})", .{ self.rank(), t.len, Tensor.MAX_RANK });
```
gwenzek, line 2030: why ? If I read the code correctly the assert is correct

zml/tokenizer/hftokenizers/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "zml_tokenizer_hftokenizers"
+version = "0.1.0"
```
gwenzek, line 3: is the Cargo stuff needed ?

I don't see bazel using the Cargo.toml file, is it read somehow ?

zml/tokenizer/hftokenizers/main.zig
```diff
@@ -0,0 +1,28 @@
+const std = @import("std");
+const c = @import("c");
+const HFTokenizers = @import("hftokenizers").HFTokenizers;
+
+pub fn main() !void {
+    const tokenizer = HFTokenizers.init("/private/var/tmp/_bazel_steeve/a67b810d44f2a673ebbd5bab86ccd5cc/external/zml~~huggingface~Meta-Llama-3.1-8B-Instruct/tokenizer.json");
```
gwenzek, line 6: this can probably be removed from master

zml/tokenizer/main.zig
```diff
@@ -0,0 +1,24 @@
+const std = @import("std");
+const tokenizer = @import("zml/tokenizer");
+
+pub fn main() !void {
+    const model1 = "/Users/steeve/Downloads/poolside.sp.pb";
+    _ = model1; // autofix
```
gwenzek: same file can be dropped

zml/tokenizer/sentencepiece/main.zig
gwenzek, line 1: feels this was a draft of smth

zml/tokenizer/sentencepiece/BUILD.bazel
```diff
@@ -0,0 +1,35 @@
+load("@rules_zig//zig:defs.bzl", "zig_library")
+load("@zml//bazel:zig.bzl", "zig_cc_binary")
+load("//bazel:swig.bzl", "swig_cc_library")
+
+swig_cc_library(
```
gwenzek, line 5: why do we need swig ? are we making a python module ?

zml/tokenizer/sentencepiece/sentencepiece2.zig
gwenzek, line 1: this looks like a lot like sentencepiece.zig do we need both ?

examples/llama/main.zig
```diff
@@ -1,156 +1,228 @@
...
+    std.debug.print("\n", .{});
+    log.info("✅ Generated {d} tokens in {:.3}s: {d:.3}tok/s", .{ num_tokens_generated, duration, speed });
     return output.toOwnedSlice();
 }
 
+const params = clap.parseParamsComptime(
+    \\--help                    print this help
+    \\--prompt <STRING>         the prompt
+    \\--model-config <PATH>     config.json path
+    \\--model-name <STRING>     model name
```
gwenzek, line 171: I don't think we need the model-name

examples/llama/test.zig
```diff
@@ -102,25 +100,27 @@ fn testImplementation(
...
-        try zml.testing.testLayer(
-            platform,
-            buffer_store,
-            "layers.0.self_attn",
-            SelfAttnPrefill{ .inner = llama.model.layers[0].self_attn },
-            .{ .inner = llama_weights.model.layers[0].self_attn },
-            1e-3,
-        );
-    }
+    // needs rework now that we don't support null tensors anymore
```
gwenzek, line 103: what's up with  those tests ?

ffi/zig_allocator.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <stdlib.h>
+
+typedef struct
+{
```
gwenzek, line 6: who is using that ?

examples/llama/main.zig
```diff
@@ -1,156 +1,228 @@
...
-
 // set this to false to disable the verbose logging
-const show_mlir = true;
+const show_mlir = false;
 
 pub const std_options = .{
     .log_level = .warn,
     .log_scope_levels = &[_]std.log.ScopeLevel{
-        .{ .scope = .zml_module, .level = if (show_mlir) .debug else .warn },
+        .{ .scope = .@"zml/module", .level = if (show_mlir) .debug else .warn },
```
gwenzek, line 26: you don't need such complicated log levels anymore, just .info for the whole thing should be enough

examples/llama/main.zig
```diff
@@ -1,156 +1,228 @@
...
         .{ .scope = .llama, .level = .info },
     },
-    .logFn = asynk.logFn,
 };
 
+pub fn tokenize(allocator: std.mem.Allocator, tokenizer: zml.tokenizer.Tokenizer, config: LlamaLM.Config, prompt: []const u8) ![]u32 {
```
gwenzek, line 31: maybe name tokenizerPrompt or smth cause this is doing more than just tokenizing. Also this feels very llama3-instruct specific. Are we fine with that ?

examples/llama/main.zig
```diff
@@ -1,156 +1,228 @@
...
+    log.info("✅ Generated {d} tokens in {:.3}s: {d:.3}tok/s", .{ num_tokens_generated, duration, speed });
     return output.toOwnedSlice();
 }
 
+const params = clap.parseParamsComptime(
+    \\--help                    print this help
+    \\--prompt <STRING>         the prompt
+    \\--model-config <PATH>     config.json path
+    \\--model-name <STRING>     model name
+    \\--model-weights <PATH>    model weights path
```
gwenzek, line 172: the model prefix make it harder to type the cli, and the code harder to read because of the `args.@"model-tokenizer"`. not sure it's worth it.

examples/llama/main.zig
```diff
@@ -159,90 +231,101 @@ pub fn asyncMain() !void {
...
-    const model_file = cli_args.model;
-
-    var arena_state = std.heap.ArenaAllocator.init(allocator);
-    defer arena_state.deinit();
-    const model_arena = arena_state.allocator();
-
-    const create_opts = try std.json.parseFromSliceLeaky(zml.Platform.CreateOptions, model_arena, cli_args.create_options, .{});
-    const platform = context.autoPlatform(create_opts).withCompilationOptions(compilation_options);
+    // initialize ZML platform with optional create options
+    // eg: --create-options='{"cuda":{"allocator":{"bfc":{"memory_fraction": 0.99}}}}'
```
gwenzek, line 235: it this working ? I'm not seeing the needed changes in platform.zig

zml/zml.zig
```diff
@@ -29,7 +29,8 @@ pub const pjrt = @import("pjrtx.zig");
 pub const testing = @import("testing.zig");
 pub const torch = @import("torch.zig");
 
-pub const tokenizer = @import("tokenizer.zig");
+// pub const tokenizer = @import("tokenizer.zig");
+pub const tokenizer = @import("zml/tokenizer");
```
gwenzek, line 33: could we move "tokenizer" to it's own library so that not all of zml requires a Rust toolchain ?

examples/llama/main.zig
```diff
@@ -1,156 +1,228 @@
...
+    const config = blk: {
+        if (res.args.@"model-config") |config_json_path| {
+            var config_json_file = try asynk.File.open(config_json_path, .{ .mode = .read_only });
+            defer config_json_file.close() catch unreachable;
+            var reader = std.json.reader(allocator, config_json_file.reader());
+            defer reader.deinit();
+            const config_obj = try std.json.parseFromTokenSourceLeaky(llama.LlamaLM.Config, allocator, &reader, .{ .ignore_unknown_fields = true });
+            break :blk config_obj;
+        } else {
+            log.err("Missing --model-config", .{});
```
gwenzek, line 221: clap doesn't support optional flags ?

zml/hostbuffer.zig
renerocksai: 👍 great suggestion!

zml/tokenizer/sentencepiece/main.zig
renerocksai: it's also in the build.bazel - so it might be good to keep it here to have a standalone smoke test of the lib w/o needing llama or sth?

zml/tokenizer/sentencepiece/sentencepiece2.zig
renerocksai: at least sentencepiece2.zig is not referenced in any build.bazel, so I remove it

examples/llama/main.zig
renerocksai: i'm removing it, good catch

examples/llama/test.zig
renerocksai: idk. but the bottom portion does not work anymore. if anyone cares, he can fix it. if not, I'll remove it. the test needs activations that are not provided (not downloaded from anywhere) anyway so they only make sense to the author as of now.

renerocksai: i mean now prefill and forward are the same, a dedicated prefill test isn't really needed imho. so i removed the commented out test

examples/llama/main.zig
renerocksai: ok

examples/llama/main.zig
renerocksai: i'll rename it. we only have llama 3 Instruct models in the build.bazel, so it's not sooooo bad it's called unconditionally.

examples/llama/main.zig
renerocksai: it isn't here. we don't have many groups of params. so I'll drop it.

examples/llama/main.zig
renerocksai: it ist. there are create opts in platform.zig and autoPlatform takes them

examples/llama/main.zig
renerocksai: all flags are zig optionals in clap. wdym?

gwenzek: I meant it doesn't support required flags

gwenzek:
closed in favor of #162 

async part is held up for later, we aren't sure about the api we want

PR #173, opened 2025-01-28
Restore tinyllama models

gwenzek:
This is a follow up on #162 

* restore homemade tokenizer, integrate it with the new zml/tokenizer library. This will only be used for working with tinyllama, I could also make it work to support GGUF if needed (couple lines of code).
* this allows us to use the original tinyllama files from karpathy repo which generates actual sentences. This is important when working on modeling because subtle bugs can degrade the output quality, which is not detectable if the output is crap to begin with.
* fix casing of zml/tokenizer api: fromFile, tokenToId. Let's stick to the Zig style guide. Function written in Rust follows rust convention, but the Zig bindings should follow Zig convention.
* make tokenizer/main.zig a proper tokenizer test tool reusing llama/test_tokenizer.zig 

~~use native_binary instead of cc_binary~~
Reverted because apparently native_binary break the runfiles and therefore the cuda sandbox


zml/tokenizer/homemade.zig
```diff
@@ -0,0 +1,1215 @@
...
+            mergeable[num_tokens] = if (token == self.special_tokens.hard_space)
+                .hard_space
+            else
+                .idk;
+        }
+
+        var stable_prefix: usize = 0;
+        var stable_off: usize = 0;
+        while (true) {
+            // This code is a bit overcomplicated cause I'm abstracting over two algorithms:
```
gwenzek, line 175: this is a restore of previous tokenizer.zig we have plus:
* a disclaimer at the top 
* this note in the middle
* binding to the encoder/decoder api

zml/tokenizer/homemade.zig
```diff
@@ -0,0 +1,1215 @@
...
+    try tokenizer.addToken(3.5, "world");
+
+    try testing.expect(tokenizer.lookup("hello") == 0);
+    try testing.expect(tokenizer.lookup("world") == 1);
+
+    // TODO: test Tokenizer.decode, Tokenizer.encode, Tokenizer.readTokenInto
+}
+
+pub const Encoder = struct {
+    inner: *Tokenizer,
```
gwenzek, line 407: Encoder struct for compat

zml/tokenizer/homemade.zig
```diff
@@ -0,0 +1,1215 @@
...
+        return res;
+    }
+
+    pub fn ids(self: *const Encoder) []const u32 {
+        return self.current_ids;
+    }
+};
+
+pub const Decoder = struct {
+    const StringBuffer = std.BoundedArray(u8, 128);
```
gwenzek, line 447: decoder struct for compat

Corendos (approved):

PR #177, opened 2025-01-29
Qwen Example

renerocksai:
WIP: Provide Qwen2.5-7B as example.

```bash
$ cd examples
$ bazel run //llama:Qwen-2.5-7B
```


gwenzek:
can you use the Instruct version so it works better with the llama instruct prompt we use ?

PR #180, opened 2025-02-03
pjrt,zml: add memory bindings

steeve:
This preliminary PR binds PJRT memory endpoints and adds them to `zml.Buffer`.

A follow up PR will properly integrate it inside `zml.Buffer`

### DIFF
--- a/pjrt/pjrt.zig
+++ b/pjrt/pjrt.zig
@@ -142,15 +142,21 @@ pub const Api = struct {
         };
     }
 
-    pub fn stablehloCurrentVersion(self: *const Api, buf: []u8) ?[]u8 {
+    pub fn stablehloCurrentVersion(self: *const Api) ?[]const u8 {
+        const state = struct {
+            var buf: [32]u8 = undefined;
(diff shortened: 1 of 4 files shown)


pjrt/pjrt.zig
```diff
@@ -377,7 +410,9 @@ pub const Client = opaque {
         element_type: BufferType,
         layout: MemoryLayout,
         device: *const Device,
-        on_delete_callback: ?*const fn (device_buffer_ptr: ?*anyopaque, ctx: ?*anyopaque) callconv(.C) void = null,
+        on_delete_callback: *const fn (device_buffer_ptr: ?*anyopaque, ctx: ?*anyopaque) callconv(.C) void = &struct {
```
gwenzek, line 413: why do you put  a callback that does nothing rather than a null ptr ?

steeve: see https://github.com/zml/zml/pull/180/files#diff-a420f288740d7de49c3cd0b14b3022be15d6d16b4c606172d16b578b726fd5ccL80

PR #185, opened 2025-02-10
pjrt: profiler support std writer API

gwenzek:
expose a more low-level function to customize where to write profile reports


hugomano (approved):
Thanks!

PR #186, opened 2025-02-10
zml: FnCache should handle non-tensor values for less confusing behavior

gwenzek:
Currently if you used the FnCache (through zml.call) on a function with non tensor fields you would get a struct with undefined memory.
This wasn't nice. So now we copy the value of the function returned by the first zml.call, and use this as base for future calls, reusing non-tensor fields, and adding correct mlir.values to the previous tensors.
This is easy because now zml.call only supports fixed size structs (no slice/pointer)


zml/module.zig
```diff
@@ -114,12 +115,15 @@ pub const CompilationContext = struct {
         var mlir_ctx = mlir.Context.initWithRegistry(mlir_registry, false) catch unreachable;
         mlir_ctx.loadAllAvailableDialects();
 
-        // Too long module names create too long file paths.
-        const name = full_name[0..@min(128, full_name.len)];
+        // Too long module names create too long file paths and files failed to create.
+        // * leave half of the space for parent folder and XLA generated filename,
+        // * leave 17 bytes for the module hash (16 + 1 for underscode).
```
Corendos, line 120:
```suggestion
        // * leave 17 bytes for the module hash (16 + 1 for underscore).
```

zml/module.zig
```diff
@@ -639,46 +646,51 @@ pub const CompilationContext = struct {
...
 
-        const values = arena.alloc(mlir.Value, function.num_args) catch unreachable;
+        const values = try arena.alloc(mlir.Value, function.num_args);
         self.extractValues(&args, values);
 
-        const donations = arena.alloc(Tensor._Donation, function.num_args) catch unreachable;
+        const donations = try arena.alloc(Tensor._Donation, function.num_args);
         meta.collectBuf(struct {
             pub fn cb(ctx: *const CompilationContext, x: Tensor) Tensor._Donation {
                 return ctx.getValueAndDonation(x)[1];
```
Corendos, line 696: We should probably clean that at some point because I see a lot of potential leaks

Corendos (approved):

zml/module.zig
gwenzek: I've doubled check the logic, allocations are split in two: memory for the caller which is part of the result and memory which is temporary because copied by MLIR.

Corendos: The fact that we use `self.allocator()` instead of `self._arena.allocator()` makes that unclear without having the context. But yes, it seems it's okay

PR #189, opened 2025-02-12
zig 0.14.0

gwenzek:


steeve:
LFG

third_party/modules/zig-protobuf/20250213.0-5304067/MODULE.bazel
gwenzek, line 1: @monadbobo

this is how we deal with third-party deps. It's a bit verbose but allows to add patches when needed

hugomano (approved):

gwenzek:
I've also bumped ZLS

PR #149, opened 2025-01-08
[WIP] Add ModernBERT implementation in examples

vctrmn:
# Description

This is a work-in-progress implementation of ModernBERT. ModernBERT is an optimized variant of BERT that includes:
- GeGLU activation function instead of GeLU
- Improved attention mechanism with local and global attention patterns
- RoPE (Rotary Position Embeddings) for better position encoding
- ...

References:
https://huggingface.co/blog/modernbert
https://huggingface.co/docs/transformers/main/en/model_doc/modernbert
https://github.com/AnswerDotAI/ModernBERT
https://github.com/huggingface/transformers/blob/main/src/transformers/models/modernbert/modular_modernbert.py

## TODO

- [x] Embeddings layer with layer normalization
- [x] Attention mechanisms (both local and global variants)
- [x] MLP with GeGLU activation
- [x] Basic test implementation
- [ ] Create ModernBertOptions struct to replace hardcoded constants:
- [ ] Implement Flash Attention 2 optimization for better throughput
- [ ] Create geglu function in tensor.zig
- [ ] Handle layer 0 special case using a union type for both "void" and zml.nn.LayerNorm cases
- [ ] Implement proper position_ids handling (currently using null)
- [ ] Handle global and local attention properly in ModernBertEncoderLayer
- [ ] Complete layers field in ModernBertModel (currently commented out)
- [ ] ... 

## Testing

Basic testing has been implemented for:
- Embeddings layer
- Attention mechanisms (both local and global)
- MLP layer
- Layer normalization
- Full encoder layer (with only local attention mechanism + without handling layer 0)

Current test tolerances range from 1e-6 to 2e-3 depending on the layer.

```bazel run -c opt //modernbert:test-implementation```

__Feedback and suggestions are welcome !__

PR #192, opened 2025-02-20
examples: add modernBERT

gwenzek:
Modern bert implementation from Victor: @vctrmn

Tested with https://huggingface.co/answerdotai/ModernBERT-base
I started from his PR #149 and cleaned up a bit the code to make it more what I consider "idiomatic ZML" even if it's still a bit in flux since we are so young.


examples/modernbert/test.zig
```diff
@@ -0,0 +1,238 @@
...
+        1e-3,
+    );
+
+    try zml.testing.testLayer(
+        compute_platform,
+        activations,
+        "model.head",
+        model.head,
+        model_weights.head,
+        0.1, // TODO: too high tolerance
```
renerocksai, line 236: Is this a problem? Or is it actually reasonable?

renerocksai (approved):
LGTM; I only added checking for null of cmdline args --model and --activations (in test target) as a just-for-the-sake-of-it CLI behavior improvement. See my last 2 commits.

Maybe it's addressed, but the tolerance in the test (see comment) has a "TODO". I tested the model, though, and it produces sensible results, so I am not sure this TODO needs to remain.

renerocksai:
BTW great work @gwenzek and @vctrmn! That was a joy to review! 👍

PR #204, opened 2025-03-18
remove usingnamespace from mlir

gwenzek:


zml/ops.zig
```diff
@@ -248,7 +240,8 @@ pub fn reduceWindow(
             .{ "window_strides", .dense(ctx.mlirCtx(), .i64, opts.window_strides) },
             .{ "base_dilations", .dense(ctx.mlirCtx(), .i64, opts.base_dilations) },
             .{ "window_dilations", .dense(ctx.mlirCtx(), .i64, opts.window_dilations) },
-            .{ "padding", .denseElements(ctx.mlirCtx(), &.{ @intCast(opts.padding.len), 2 }, .i64, opts.padding) },
+            // Cast the [][2]i64 to []i64 (safe)
+            .{ "padding", .denseElements(ctx.mlirCtx(), &.{ @intCast(opts.padding.len), 2 }, .i64, @ptrCast(opts.padding)) },
```
Corendos, line 244: I don't think that's safe, pointers are compatible but length are not, no ?
`[]i64` should have twice as many elements at `[][2]i64`

Corendos (approved):

PR #212, opened 2025-03-25
async: detect xev most preferable backend to use

hugomano:
Since change allow use to leverage io_uring instead of xev if the backend is supported.


Corendos (approved):
shipit

PR #237, opened 2025-05-22
zml: remove compile time fields from zml.Bufferized

gwenzek:
this remove the need to pass `undefined` to `exe.call` when the corresponding `.forward` function expects integer/float arguments that have been inlined in the executable.

This was not needed and could just create confusion.
Now you may end up having to pass void to `exe.call` if the executable doesn't have any runtime argument,
but it's still better than passing `undefined`.

The only place that was using the non-buffer parts of a `zml.Bufferized` object was the infamous `zml.testing.compileAndCall`.
So I had to introduce a `zml.testing.BufferizedWithArgs` which is only used for this function.
I put it under the `zml.testing` namespace to make it clear it's only for this testing utility.
Note that because `compileAndCall` is strongly typed, you generally don't need to explicitly mention `BufferizedWithArgs` 
and the can just write the arguments in place and get help from ZLS.

### DIFF
--- a/zml/exe.zig
+++ b/zml/exe.zig
@@ -231,6 +231,35 @@ pub const BaseExe = struct {
         }
     }
 
+    pub fn _unsafeAssignResults(self: BaseExe, T: type, result: *T) void {
+        const LocalContext = struct {
+            index: u32,
+            platform: Platform,
+            outputs: []const [*]*pjrt.Buffer,
+            output_shapes: []Shape,
+        };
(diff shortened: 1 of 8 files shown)


Corendos (approved):
Just hope it's going to be compatible with #251 👀

gwenzek:
Good point about #251, I've added union support in MapRestrict, I'll let you do it for Map.
But I've added a testcase for Map that will fail once you add union support.

PR #244, opened 2025-05-30
zml: custom call v2

hugomano:
This PR introduce new APIs for custom call with multi target compatibilities and commands buffers for graph capture. Also fix some PjRT API bindings and code smells.


gwenzek:
This now conflicts with #248

PR #248, opened 2025-06-04
zml/buffer: various improvements

Corendos:
* `Buffer.from` is blocking by default
* `Buffer.from` takes options to control the loading This includes awaiting for the buffer or to which memory the buffer is copied

### DIFF
--- a/zml/buffer.zig
+++ b/zml/buffer.zig
@@ -51,8 +51,13 @@ pub const Buffer = struct {
     pub const MAX_NUM_SHARDS: u8 = Platform.MAX_NUM_DEVICES;
     pub const Shards = std.BoundedArray(*pjrt.Buffer, MAX_NUM_SHARDS);
 
+    pub const FromOptions = struct {
+        wait: bool = true,
+        memory: ?pjrt.Memory.Kind = null,
+    };
+
     /// Copies the content of the given buffer from host memory to the accelerator memory.
-    pub fn from(platform: Platform, host_buffer: HostBuffer) !Buffer {
+    pub fn from(platform: Platform, host_buffer: HostBuffer, opts: FromOptions) !Buffer {
         var res: Buffer = .{
             ._api = platform.pjrt_api,
             ._shape = host_buffer.shape(),
@@ -73,7 +78,6 @@ pub const Buffer = struct {
         const buffer_type = bufferTypeFromDtype(host_buffer.shape().dtype());
         const byte_strides = host_buffer.strides();
 
-        var frames: std.BoundedArray(asynk.Frame(pjrt.Client.bufferFromHostBuffer), MAX_NUM_SHARDS) = .{};
         const devices = platform.getDevices();
         for (0..n_partitions) |i| {
             // If no sharding if found, the given buffer is replicated on all devices.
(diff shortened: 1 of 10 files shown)


hugomano (approved):
Thanks!

zml/buffer.zig
```diff
@@ -122,20 +147,40 @@ pub const Buffer = struct {
...
-        return from(platform, HostBuffer.fromBytes(sh, data));
+        return from(platform, HostBuffer.fromBytes(sh, data), .{});
     }
 
     /// Copies the given Zig array to the accelerator memory and
     /// return a Buffer using the array shape.
     pub fn fromArray(platform: Platform, arr: anytype) !Buffer {
         const host_buffer = HostBuffer.fromArray(&arr);
-        return try from(platform, host_buffer);
+        return try from(platform, host_buffer, .{});
```
gwenzek, line 163: fromArray should wait by default, because the arrays are typically on the stack.

zml/buffer.zig
```diff
@@ -176,7 +221,7 @@ pub const Buffer = struct {
                 ._strides = @splat(0),
                 ._data = x.constSlice().ptr,
             };
-            return try from(platform, host_buffer);
+            return try from(platform, host_buffer, .{});
```
gwenzek, line 224: same

zml/buffer.zig
```diff
@@ -199,7 +244,7 @@ pub const Buffer = struct {
             else => unreachable,
         }
         const host_buffer: HostBuffer = .{ ._shape = shape_, ._strides = strides, ._data = &bytes };
-        return try from(platform, host_buffer);
+        return try from(platform, host_buffer, .{});
```
gwenzek, line 247: same.

zml/hostbuffer.zig
```diff
@@ -158,7 +158,7 @@ pub const HostBuffer = struct {
 
     /// Copies this HostBuffer to the given accelerator.
     pub fn toDevice(self: HostBuffer, platform_: Platform) !Buffer {
```
gwenzek, line 160: we probably want a toDeviceOpt too.

zml/buffer.zig
```diff
@@ -51,8 +51,13 @@ pub const Buffer = struct {
     pub const MAX_NUM_SHARDS: u8 = Platform.MAX_NUM_DEVICES;
     pub const Shards = std.BoundedArray(*pjrt.Buffer, MAX_NUM_SHARDS);
 
+    pub const FromOptions = struct {
+        wait: bool = true,
+        memory: ?pjrt.Memory.Kind = null,
```
gwenzek, line 56: can we replace the optional by the actual default value ? `.device` IIUC

zml/buffer.zig
```diff
@@ -122,20 +147,40 @@ pub const Buffer = struct {
...
         const host_buffer = HostBuffer.fromArray(&arr);
-        return try from(platform, host_buffer);
+        return try from(platform, host_buffer, .{});
+    }
+
+    /// Copies the given Zig slice to the accelerator memory and
+    /// return a Buffer with the given dimensions.
+    pub fn fromSliceOpts(platform: Platform, dimz: anytype, s: anytype, opts: FromOptions) !Buffer {
+        const sh = Shape.init(dimz, DataType.fromSliceElementType(s));
+        return from(platform, HostBuffer.fromBytes(sh, std.mem.sliceAsBytes(s)), opts);
```
gwenzek, line 170: what happens here if you create a .host slice ? does PJRT makes a copy ?

zml/buffer.zig
Corendos: I have mixed feelings about it. I don't know what a sane default would be. For now it's true that on platform we support, it's copied to `.device` memory but will that always be the case ?

zml/buffer.zig
Corendos: I just tested, yes it does

gwenzek (approved):

PR #249, opened 2025-06-04
zml/tensor.zig: `dynamicSlice1d` accepts tags

Corendos:


hugomano (approved):

Corendos:
Ah, ok
Might be a footgun but I'll close the PR then

PR #253, opened 2025-06-04
zml/ops: added customCall

Corendos:


gwenzek (changes requested):
Can you add a test or at least an example usage ?
Like can you call the cudnn sdpa with it ?

Corendos:
Cudnn sdpa is probably too much work but I'll find something

gwenzek (approved):

gwenzek:
the `print` could be a good first example, but approving to not block

PR #269, opened 2025-06-20
runtimes/cuda: downgrade cuda and cudnn

Corendos:
This commit reverts part of https://github.com/zml/zml/pull/238/files This is required because XLA has a strong dependencies on CUDA 12.8 and upgrading to 12.9 is impossible due to https://github.com/NVIDIA/cccl/issues/4967


hugomano (approved):
Thanks, good catch!

PR #278, opened 2025-07-15
runtimes/tpu: sandbox TPU PJRT Plugin

cerisier:
It has no dependencies so fairly straighforward


runtimes/tpu/tpu.zig
```diff
@@ -37,5 +40,18 @@ pub fn load() !*const pjrt.Api {
         return error.Unavailable;
     }
 
-    return try asynk.callBlocking(pjrt.Api.loadFrom, .{"libpjrt_tpu.so"});
+    var arena = std.heap.ArenaAllocator.init(std.heap.c_allocator);
```
gwenzek, line 43: I'd rather avoid std.heap.c_allocator. std.heap.smp_allocator is just as fast and doesn't depend on libc.

steeve (approved):

PR #279, opened 2025-07-15
runtimes/cpu: sandbox CPU PJRT Plugin

cerisier:
Same as TPU, simplified because there are no other NEEDED dependencies.

### DIFF
--- a/runtimes/cpu/cpu.zig
+++ b/runtimes/cpu/cpu.zig
@@ -3,6 +3,12 @@ const builtin = @import("builtin");
 const asynk = @import("async");
 const c = @import("c");
 const pjrt = @import("pjrt");
+const bazel_builtin = @import("bazel_builtin");
+const std = @import("std");
+const stdx = @import("stdx");
+const runfiles = @import("runfiles");
+
+const log = std.log.scoped(.@"zml/runtime/cpu");
 
 pub fn isEnabled() bool {
     return @hasDecl(c, "ZML_RUNTIME_CPU");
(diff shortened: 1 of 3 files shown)


runtimes/cpu/cpu.zig
```diff
@@ -13,10 +19,25 @@ pub fn load() !*const pjrt.Api {
         return error.Unavailable;
     }
 
+    var arena = std.heap.ArenaAllocator.init(std.heap.c_allocator);
```
gwenzek, line 22: I'd rather avoid std.heap.c_allocator. `std.heap.smp_allocator` is just as fast and doesn't depend on libc.

runtimes/cpu/cpu.zig
```diff
@@ -13,10 +19,25 @@ pub fn load() !*const pjrt.Api {
...
+    defer arena.deinit();
+
+    var r_ = try runfiles.Runfiles.create(.{ .allocator = arena.allocator() }) orelse {
+        std.debug.panic("Unable to create Runfiles isntance.", .{});
```
gwenzek, line 26: Could we have another way to pass `zml/runtimes/`, maybe through a flag ?

runtimes/cpu/cpu.zig
```diff
@@ -13,10 +19,25 @@ pub fn load() !*const pjrt.Api {
...
+    };
+    const source_repo = bazel_builtin.current_repository;
+    const r = r_.withSourceRepo(source_repo);
+    const cuda_data_dir = (try r.rlocationAlloc(arena.allocator(), "zml/runtimes/cpu/sandbox/lib")).?;
```
gwenzek, line 30: cpu_data_dir

gwenzek (approved):

runtimes/cpu/cpu.zig
cerisier: Very doable yes, I'll check with @steeve 
Actually could be handy for all other runtimes as well when testing debug versions.

PR #292, opened 2025-07-31
examples: Mixtral

gwenzek:
to try it:
* download mixtral model first with: `huggingface-cli download mistralai/Mixtral-8x7B-Instruct-v0.1 --exclude='*.pt' --local-dir=$HOME/models/Mixtral-8x7B-Instruct-v0.1`
* cd examples; `bazel run --config=debug //mixtral -- --hf-model-path=~/models/Mixtral-8x7B-Instruct-v0.1/`

This includes:
* high level bindings for `AsyncHostToDeviceTransferManager`
  This was needed because I didn't like how Mixtral stores the experts has separate buffers. 
  This implementation creates one `TransferManager` per layer and use it to load the weights in one buffer.
  There is more work to do to make it sharding aware, currently I've hardcoded the sharding in Mixtral code, which isn't ideal.
 * zml.nn.mixtureOfExperts with 3 implementations that the user can select from with different precision/throughput tradeoffs
 * `Tensor.gather` refactoring using a syntax similar `Tensor.scatter`
 * `Tensor.topK` has a new syntax that forces you to rename the axes when using tagging. In practice most of the code was already doing this because it's error prone otherwise.


gwenzek:
Replaced by #329 , sorry Mistral

PR #319, opened 2025-08-29
zml: custom call api

gwenzek:
Resuming work from @hugomano #244 
This is still a bit experimental.

The core API is here, but there are some caveats.

ZML should help with copying buffers to host memory (which is often needed) but I'm not sure the current implementation using PJRT is good:
- does it allow graph capture
- what's the overhead ?
We should probably use directly cuMemcpy here instead of going through PJRT.
Current implem should work with all plugins though.

This fix `tensor.print()`

I've also made a slightly more interesting examples/custom_call that manually calls a cuda kernel. (there are other way to achieve this but this shows you can do whatever you want with `zml.customCall`)


zml/zml.zig
```diff
@@ -6,11 +6,13 @@
 // Namespaces
 const std = @import("std");
 
+pub const platform_specific = @import("c");
```
gwenzek, line 9: exposes cuda symbols to user with a scary name.

PR #322, opened 2025-09-05
zml: loadById

gwenzek:
We already assign a unique tensor id to each tensor object created by a BufferStore.
But we weren't using the id to load find the corresponding buffer in the store, instead relying on the name.

This is allow to decouple the shape of the model in the struct vs the shape of the model inside the safetensor file.
I also moved the id logic out of Tensor and into BufferStore.

I also updated llama/main.zig to use the new logic. 

Bonus:
* fix to `zml.call` tag hashing leading to different MLIR for every recompile.
* expose  Cpu device count to control the number of CPU "devices"

### DIFF
--- a/zml/aio.zig
+++ b/zml/aio.zig
@@ -111,7 +111,7 @@ pub const BufferStore = struct {
     }
 
     pub fn loadBufferById(self: BufferStore, x: zml.Tensor, platform: zml.Platform) !zml.Buffer {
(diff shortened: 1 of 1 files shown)


gwenzek:
@hugomano I believe that `store.loadModelById` could work very well with the transfer manager cause the store knows how to read buffers and the model struct say how they should be sharded.

PR #324, opened 2025-09-15
zml: remove old writer interface APIs

gwenzek:


Corendos (approved):
Thanks

I've skimmed over the mlir part as it will have an overhaul soon

PR #348, opened 2025-10-17
all: provide -fllvm to all targets that requires it

cerisier:
Until zig x86 backend is fixed.


gwenzek:
Did you try with Zig 0.15.2 ?

cerisier:
Yes, doesn't solve anything :(

PR #372, opened 2026-01-07
remove hack no longer needed in Buffer.uninitialized

gwenzek:
This has been fixed upstream, we can remove the hack.


hugomano:
Done in zml/v2 now merged in master.

PR #405, opened 2026-03-09
docs: update bazel paths and huggingface cli command

carlos-gn:


Corendos:
Hi @carlos-gn, sorry for the delay. We updated the doc with the V2 release so I'll close this.

Thanks for the time you spent though 😁

PR #416, opened 2026-03-18
zml/tokenizer: add IREE tokenizer

Finistere:
Adding the [IREE tokenizer](https://github.com/iree-org/iree/blob/main/runtime/src/iree/tokenizer/README.md).Being made from the ground up with streaming (de-)tokenization in mind, I've extended the APIs to expose a `std.Io.Writer` for the tokenization.

I've quickly benchmarked IREE's tokenizer vs HuggingFace's one with a request I retrieved from a coding agent (opencode) in the middle of a session. The request body was 360KB encoded in 95K tokens. I also did another test with a similarly sized request with the same results as shown hereafter.

I've tested two cases:

- oneshot: the template is rendered all at once and only then we encode the whole string.
- stream: the template rendering and encoding are pipelined to avoid the intermediate allocation.

There is a slight difference in the generated tokens between HuggingFace and IREE. Some tokens like `introdu` were encoded as `i`, `n`, `t`, etc. Decoding them returns the same text. It's very close in terms of total token count:

- hf: 95779
- iree/stream: 95705
- iree/oneshot: 95587

| Model | IREE avg (oneshot / stream) ms | HuggingFace avg (ms) | Uplift vs HF (oneshot / stream) |
|---|---:|---:|---|
| liquidai/LFM2.5-1.2B-Instruct | 8.5 / 7.3 | 40.5 | 4.8x / 5.5x |
| Qwen/Qwen3.5-9B | 6.0 / 8.2 | 59.4 | 9.9x / 7.2x |
| mistralai/Ministral-3-3B-Reasoning-2512 | 7.4 / 6.9 | 59.2 | 8x / 8.6x |

It's unclear why the Qwen's chat template is somehow slower in the stream case, maybe because there are more writes. Adding non-empty buffer in the interface helps a bit, but tends to slow down mistral case. So well


steeve (approved):

stellaraccident:
Nice!

PR #448, opened 2026-04-02
zml/attention: Move zml.attention.attention.* to zml.attention.*

Finistere:
Avoids the unnecessary nesting.


gwenzek (approved):

Finistere:
I didn't have in mind how paged attention is structured, will try to come up with something better than this

PR #449, opened 2026-04-02
Moe backend (triton bf16 and fp8)

loupicaaa:
### Overview

**Context**: Many modern models utilize a type of Multi-Layer Perceptron (MLP) called Mixture of Experts (MoE). In an MoE architecture, the MLP is organized into discrete "experts," where specific sub-tensors are specialized to process particular linguistic patterns or elements. During inference, the hidden states of each token are routed to the most relevant experts, ensuring that matrix multiplications (matmul) are performed only on the active parameters.

**Problem Statement**: Due to the sparse and dynamic nature of MoE—where token assignment is determined at runtime via routing, treating an MoE layer as a standard matmul results in significant performance degradation. Standard kernels are not optimized for the irregular memory access patterns inherent in sparse routing.

**Approach**: MoE is implemented as a core primitive within the ZML operation set. The Generic MoE Backend is designed to support various specialized kernels tailored to specific hardware and weight precisions. The primary implementation currently features a Triton MoE backend optimized for NVIDIA Hopper architecture using BF16 and FP8 weights. This implementation is utilized in the Qwen 3.5 MoE model.

### Features

**Generic MoE Backend** (zml/moe/moe.zig)

- Extensibility: Designed to integrate specialized MoE backends (e.g., Triton).
- Dispatch: Select the optimal backend based on hardware and precision.
- Infrastructure: Implements essential utilities and structures for registration, parameter loading, and metadata management.
- Execution: The forwardMoe() function acts as the entry point to invoke the appropriate backend.

**Triton MoE Backend** (zml/moe/triton.zig)

- Compilation: Generates TTIR (Triton Intermediate Representation) at compile time.
- XLA Integration: Leverages Triton's custom calls exposed via XLA.
- Execution Flow: 

2. Determine token distribution: Use naive assignment if the distribution is sparse; otherwise, sort and align tokens for efficiency.
3. Quantization (Initial): Quantize activations if using FP8 weights.
4. Fused Kernel (Gate/Up): Execute the fused MoE kernel for the gate and up-projection layers.
5. Post-Processing: Split the output and apply the activation function.
6. Quantization (Intermediate): Re-quantize activations for the next stage if required by the FP8 path.
7. Fused Kernel (Down): Execute the fused MoE kernel for the down-projection layer.
8. Aggregation: Sum the Top-K results for each token.

**Qwen 3.5 MoE**

- Backend Integration: Fully utilizes the Triton MoE backend.
- Precision Support: Supports both BF16 and FP8 (providing distinct paths for quantized and unquantized weights).
- FP8 Optimization: Implements expert repacking and a scaled linear forward pass for the relevant layers.


zml/zml.zig
```diff
@@ -46,6 +46,8 @@ pub const tensor = @import("tensor.zig");
 pub const Tensor = tensor.Tensor;
 pub const testing = @import("testing.zig");
 pub const tracer = @import("tracer.zig");
+pub const triton_moe = @import("moe/triton.zig");
+pub const moe = @import("moe/moe.zig");
```
Corendos: We should probably expose only `moe` in `zml`

zml/moe.zig
Corendos, line 1: The `moe.zig` file can be put directly in the `zml/` folder

zml/moe/triton.zig
```diff
@@ -0,0 +1,908 @@
...
+    return out;
+}
+
+fn i64s(v: i64) Tensor {
+    return Tensor.constant(.{ .i64 = v }).reshape(.{1});
+}
+
+fn ceilDiv(a: i64, b: i64) i64 {
+    return @divFloor(a + b - 1, b);
+}
```
Corendos: You can replace with `std.math.divCeil`

zml/moe/triton.zig
```diff
@@ -0,0 +1,908 @@
...
+    out.group_size_m = try jsonValueAsI64(cfg.get("GROUP_SIZE_M") orelse return error.MissingLaunchConfigField);
+    out.num_warps = try jsonValueAsI64(cfg.get("num_warps") orelse return error.MissingLaunchConfigField);
+    out.num_stages = try jsonValueAsI64(cfg.get("num_stages") orelse return error.MissingLaunchConfigField);
+
+    return out;
+}
+
+fn i64s(v: i64) Tensor {
+    return Tensor.constant(.{ .i64 = v }).reshape(.{1});
+}
```
Corendos: Too specific, you can replace at call site with `Tensor.constant(.init(.i64, v)).reshape(.{1})`

zml/moe/triton.zig
```diff
@@ -0,0 +1,908 @@
...
+    return try allocator.dupe(u8, config_json);
+}
+
+fn jsonValueAsI64(v: std.json.Value) !i64 {
+    return switch (v) {
+        .integer => |x| x,
+        .float => |x| @intFromFloat(x),
+        else => error.InvalidLaunchConfigValue,
+    };
+}
```
Corendos: The value can really be a float ? 😭

zml/moe/triton.zig
```diff
@@ -0,0 +1,908 @@
...
+    if (gate_up.dtype() != .bf16 and gate_up.dtype() != .f8e4m3fn) return error.UnsupportedType;
+    if (down.dtype() != .bf16 and down.dtype() != .f8e4m3fn) return error.UnsupportedType;
+    if (weights.dtype() != .f32 and weights.dtype() != .bf16) return error.UnsupportedType;
+    if (ids.dtype() != .i32) return error.UnsupportedType;
+    if (hidden.dim(.in) != gate_up.dim(.in)) return error.InvalidShape;
+    if (@rem(gate_up.dim(.out), 2) != 0) return error.InvalidShape;
+    if (down.dim(.mid) != @divFloor(gate_up.dim(.out), 2)) return error.InvalidShape;
+    if (ids.dim(.token) != hidden.dim(.token) or weights.dim(.token) != hidden.dim(.token)) return error.InvalidShape;
+    if (ids.dim(.topk) != weights.dim(.topk)) return error.InvalidShape;
+    if (gate_up.dim(.expert) != down.dim(.expert)) return error.InvalidShape;
```
Corendos: Maybe this could be factorized in a `validateInputs` function

zml/tensor.zig
```diff
@@ -1739,9 +1739,10 @@ pub const Tensor = struct {
     /// Concatenates the input Tensors along the given axis.
     pub fn concatenate(tensors: []const Tensor, axis_: anytype) Tensor {
         if (tensors.len == 1) return tensors[0];
-        stdx.debug.assert(tensors.len <= 32, "concatenate only supports up to 32 tensors, got {}", .{tensors.len});
-        var buffer: [32]*const mlir.Value = undefined;
-        std.debug.assert(tensors.len <= buffer.len);
+
+        const allocator = CompilationContext.current().arena.allocator();
+        var buffer = allocator.alloc(*const mlir.Value, tensors.len) catch std.debug.panic("concatenate failed to allocate buffer for mlir values", .{});
```
gwenzek, line 1744: Thanks

I think `@panic("OOM")` here is enough since it's generally not a user side problem, you can probably try again and it will work.

zml/moe/triton/BUILD.bazel
```diff
@@ -0,0 +1,113 @@
...
+        "--index-strategy=unsafe-best-match",
+        "--upgrade",
+        "--python-platform=x86_64-manylinux_2_37",
+    ],
+    python_version = "3.12",
+    tags = ["manual"],
+)
+
+py_library(
+    name = "fake_plugin",
```
gwenzek, line 26: can you explain what the "fake plugin" is doing ?

zml/moe/triton/BUILD.bazel
```diff
@@ -0,0 +1,113 @@
...
+cc_library(
+    name = "libpython",
+    hdrs = ["libpython.h"],
+    deps = [
+        "@python_3_12//:libpython",
+        "@python_3_12//:python_headers",
+    ],
+)
+
+zig_binary(
```
gwenzek, line 65: why do you need to call generate.py through a Zig program ?

gwenzek (commented):
wow it's a big one, thanks for all the work on plugin the triton Moe kernels.

I don't have full context here, so letting others approve wrt the overall approach.

zml/moe/triton.zig
```diff
@@ -0,0 +1,908 @@
...
+    max_num_m_blocks: usize,
+    block_size_m: usize,
+    experts_per_warp: usize,
+    hist_block: usize,
+    sort_block_size: usize,
+    sort_grid_x: usize,
+};
+
+const QuantGenerationConfig = struct {
+    kernel_family: []const u8 = "activation_quant_fp8",
```
Corendos: Not used as well

zml/moe/triton.zig
```diff
@@ -0,0 +1,908 @@
...
+        ttir_first_matmul,
+        first_generation_config,
+        max_num_tokens_padded,
+        num_assignments,
+        Shape.init(.{ .token = hidden_quant.dim(.token) * ids.dim(.topk), .out = gate_up.dim(.out) }, .bf16),
+    );
+
+    const first_flat = first_out.reshape(.{ .g = num_assignments, .out = gate_up.dim(.out) });
+
+    const gate, const up = zml.nn.splitRealImg(first_flat, .sequential);
```
Corendos: This feels weird, I think we could find a clearer way

zml/moe/triton.zig
Corendos, line 1: Some comments on the file globally:

* We should try to avoid `std.heap.c_allocator`
* TTIR generation should rely on a always running process to save triton startup time (see https://github.com/zml/zml/blob/8d3f69fa0cc767a040005bd3b94722b27321d570/zml/attention/triton.zig#L213-L237)
* We could probably factorize the `generateXXX` function by having a config union whose tag is the kernel family

examples/llm/BUILD.bazel
```diff
@@ -9,6 +9,10 @@ zig_binary(
     name = "llm",
     srcs = glob(["**/*.zig"]),
     main = "main.zig",
+    data = [
+        "//zml/moe/triton:sandbox",
+        "//zml/moe/triton:triton_kernels/config.json",
+    ],
```
Corendos, line 15: This should be done on `zml` not `llm`

examples/llm/models/qwen3_5_moe/model.zig
```diff
@@ -0,0 +1,1715 @@
...
+        k = self.k_norm.forward(k.rename(.{ .hd = .d })).rename(.{ .d = .hd });
+
+        const dtype = q.dtype();
+        const position_ids = zml.Tensor.arange(.{ .end = x.dim(.s) }, .i64)
+            .withTags(.{.s}).insertAxes(.s, .{.b}).broad(zml.Shape.init(.{ .b = x.dim(.b), .s = x.dim(.s) }, .i64))
+            .add(token_index.convert(.i64).broad(zml.Shape.init(.{ .b = x.dim(.b), .s = x.dim(.s) }, .i64)));
+
+        const cos, const sin = self.rotary_embed.getCosAndSin(position_ids, dtype);
+        q = self.rotary_embed.applyRope(q, cos, sin);
+        k = self.rotary_embed.applyRope(k, cos, sin);
```
Corendos, line 736: Why does `zml.nn.rope` doesn't work ?

zml/moe/triton/BUILD.bazel
loupicaaa: Triton compilation require to be coupled with a platform (I guess because of the jit nature of triton). But in our case we want to generate the ttir AoT so we decouple it and we don't need to link with CUDA (or whatever platform) for real compilation (it's made with XLA from the generated ttir with ops.triton)

examples/llm/models/qwen3_5_moe/model.zig
```diff
@@ -0,0 +1,1715 @@
...
+    gate_up_proj: ?zml.Tensor,
+    gate_up_proj_scale: ?zml.Tensor,
+    slice_gate_proj: []zml.Tensor,
+    slice_gate_proj_scale: []zml.Tensor,
+    slice_up_proj: []zml.Tensor,
+    slice_up_proj_scale: []zml.Tensor,
+    down_proj: ?zml.Tensor,
+    down_proj_scale: ?zml.Tensor,
+    slice_down_proj: []zml.Tensor,
+    slice_down_proj_scale: []zml.Tensor,
```
Corendos, line 868: If we never use the slice version, they shouldn't be in the model

examples/llm/models/qwen3_5_moe/model.zig
Corendos, line 1: I think we have a lot of opportunities to factorize code and find nicer patterns here. We can revisit it later, but spending some time now could be valuable.

Corendos (commented):
Left a bunch of comments 😁

zml/moe/triton/BUILD.bazel
loupicaaa: It's to avoid to embed a python interpreter

zml/moe/triton.zig
loupicaaa: Reshape was useles I removed it

PR #450, opened 2026-04-02
zml/attention: attnd

Finistere:
Simple reference implementation of attnd client as a attention backend relying on POSIX. I've tried various things with `sendmmsg`/`recvmmsg` and `io_uring` out of curiosity but they weren't worth the extra code here.

The integration as a attention backend is a bit clunky, we're still allocating a KV cache and there are a few panics here and there because the current initialization mechanism of attention backend doesn't work for attnd, but that's something for later.

### DIFF
--- a/examples/llm/main.zig
+++ b/examples/llm/main.zig
@@ -19,6 +19,7 @@ const Args = struct {
     seqlen: u32 = 2048,
     topk: u32 = 4,
     backend: ?zml.attention.attention.Backend = null,
+    attnd_ip: ?[]const u8 = null,
     single: bool = false,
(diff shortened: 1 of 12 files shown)


gwenzek (approved):
I'd like a followup later, to remove special casing of attnd init

examples/llm/main.zig
```diff
@@ -161,13 +167,25 @@ pub fn main(init: std.process.Init) !void {
     );
     defer llm_chat.deinit();
 
+    // var profiler = try platform.profiler(allocator, io, .{
```
gwenzek: remove

PR #457, opened 2026-04-10
zml/module: provide an Io to the CompilationContext

Corendos:
This might come in handy in the future if we need to do some IO in the compilation pipeline

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -81,6 +81,7 @@ pub const CompilationContext = struct {
     };
 
     allocator: std.mem.Allocator,
+    io: std.Io,
     arena: std.heap.ArenaAllocator,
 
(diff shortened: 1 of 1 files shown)


Finistere:
Is it really worth adding an `Io` in the context with a big comment explaining why it's not that simple when it only "might come handy"?
It doesn't seem obvious to me we'll need an `Io` directly in the compilation context rather than a cache backed by an `Io` for example, but you know better. 🤷

Corendos:
Good point, the reasoning is that we already have an `Allocator` and `Io` is the other "pass everywhere" struct.

Also, I needed an `Io` for the unified attention PR so I think it can be useful

hugomano (approved):

hugomano (approved):

PR #459, opened 2026-04-10
llm: remove wrong computation of num_heads

Corendos:
This confuses the attention backend because it's wrong

The real fix is to make the attention sharding aware


Corendos:
Closed in favor of #461

PR #461, opened 2026-04-10
zml/attention & zml/ops: mark fa2 as sharding aware and fix implementation

Finistere:
Sharding aware case wasn't properly handled and I marked fa2 as sharding unaware even though it is.


Corendos (approved):
The divider we get from the `CompilationContext` feels weird but we'll find a nicer solution later

steeve (approved):

PR #475, opened 2026-04-14
examples/llm: shard and async compilation for lfm

elogir:
This is not really useful since there is no gain (max LFM 2.5 weights are 2B params) but allow us to quickly test sharding on runtimes with a slow compiler. Also parallelize compilation.


Finistere (approved):

hugomano:
Closed in favor of https://github.com/zml/zml/pull/555

PR #482, opened 2026-04-17
zml/tokenizer/iree: Improve IREE tokenizer with upstream fix

Finistere:
I've reworked the Tokenizer API to actually work with all tokenizers and not just IREE and moved the `Writer` part into the IREE tokenizer. I'm now avoiding intermediate copies as much as possible now.

I'm also more defensive now to avoid the same infinite loop and put more effort into providing appropriate output buffers to IREE. It fails with empty output buffers and may return a `ResourceExhausted` if it's too small. Also added a bunch of tests.

IREE tokenizer itself had a bug which I fixed here https://github.com/iree-org/iree/pull/24150. In the meantime, I'm pointing bazel to our fork.


third_party/iree/repo.bzl
```diff
@@ -3,8 +3,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def repo():
     git_repository(
         name = "iree",
-        remote = "https://github.com/iree-org/iree.git",
-        commit = "71af3a5e41a8e265330bc693194c708cf6df4724",
+        remote = "https://github.com/zml/iree.git",
+        commit = "96151c5c077dc56c00e2b608c4f39b5553089515",
```
steeve: can you instead of pointing to the fork, create a `.patch` and apply it at repo time ?

steeve (changes requested):

third_party/iree/repo.bzl
Finistere: done

steeve (approved):

PR #490, opened 2026-04-21
ci: bump ci actions (Node.js 20 actions are deprecated)

hugomano:
Use newer versions of several GitHub Actions. Also bump used Zig version. 
`mlugg/setup-zig` is still using Node.js v20 so warning still here but only for this action. Check https://codeberg.org/mlugg/setup-zig/pulls/55 in a near future in order to update the action.


Finistere (approved):
nice

PR #496, opened 2026-04-22
zml/moe : Moe backend triton, qwen3_5_moe

loupicaaa:
### Overview
**Context**: Modern models utilize a MLP called Mixture of Experts (MoE). In a MoE, the MLP is organized into "experts," where specific sub-tensors are specialized to process particular linguistic patterns or elements. During inference, the hidden states of each token are routed to the most relevant experts, ensuring that matrix multiplications (matmul) are performed only on the active parameters.

**Problem**: Due to the sparse and dynamic nature of MoE where token assignment is determined at runtime via routing, treating an MoE layer as a standard matmul results in significant performance degradation. Standard kernels are not optimized for the irregular memory access patterns in sparse routing.

**Approach**: MoE is implemented as a primitive in the ZML operation set. The generic MoE Backend is designed to support various specialized kernels tailored to specific hardware and weight precisions. The primary implementation currently features a Triton MoE backend optimized for NVIDIA Hopper architecture using BF16 and FP8 weights. This implementation is utilized in the Qwen 3.5 MoE model.

### Features

**Generic MoE Backend (zml/moe/moe.zig)**
- Integrate specialized MoE backends (e.g., Triton).
- Select the optimal backend based on hardware and precision.
- Implements essential utilities and structures for registration, parameter loading, and metadata management.
- The forwardMoe() function acts as the entry point to invoke the appropriate backend.

**Triton MoE Backend (zml/moe/triton.zig)**
- Generates TTIR (Triton Intermediate Representation) at compile time. It uses a triton runtime that must be init with platform (initMoeBackend)
- Leverages Triton's custom calls exposed via XLA.
- Execution Flow:
2. Determine token distribution: use naive assignment if the distribution is sparse; otherwise, sort and align tokens for efficiency.
3. Quantization: quantize activations if using FP8 weights.
4. Fused Kernel: execute the fused MoE kernel for the gate and up-projection layers.
5. Post-Processing: split the output and apply the activation function.
6. Quantization : re-quantize activations for the next stage if required by the FP8 path.
7. Fused Kernel: execute the fused MoE kernel for the down-projection layer.
8. Aggregation: sum the Top-K results for each token.

**Qwen 3.5 MoE**
- Fully utilizes the Triton MoE backend.
- BF16 weights implementation (also tested with FP8 but not implemented here)

### Refactoring:
- Keep the files related to attention triton in zml/attention with other attention backends.
- generateTtir uses a specific struct to handle the ttir generation in the sandbox : 
```
         .{ .backend = backend,
            .kernel = kernel,
            .config = config, } 
```
- Introduce a zml/triton folder with the triton generic files. Contains a runtime.zig common to the triton backends (MoE and   attention).


zml/triton/runtime.zig
```diff
@@ -0,0 +1,42 @@
...
+
+        const runfiles = bazel.runfiles(bazel_builtin.current_repository) catch unreachable;
+        const sandbox_path = runfiles.rlocationAlloc(arena.allocator(), "zml/zml/triton/generation/sandbox") catch unreachable;
+
+        var map: std.process.Environ.Map = .init(arena.allocator());
+        try map.put("LD_LIBRARY_PATH", std.Io.Dir.path.join(arena.allocator(), &.{ sandbox_path.?, "lib" }) catch unreachable);
+
+        const process_mutex = try allocator.create(std.Io.Mutex);
+        errdefer allocator.destroy(process_mutex);
+        process_mutex.* = .init;
```
Corendos, line 22: Maybe put a small comment explaining why it needs to be a pointer.

zml/moe/triton/moe_compile.py
Corendos, line 1: Maybe we could share some stuff between all the generation files. `FakeTensor` for example could be shared. There are probably others as well that could be shared.

zml/attention/triton.zig
```diff
@@ -201,7 +201,13 @@ pub const GenerationConfig = union(KernelKind) {
     reduce_segments_ptr: GenerationConfigReduce,
 };
 
-fn generateTtir(allocator: std.mem.Allocator, io: std.Io, config: GenerationConfig) ![:0]const u8 {
+fn generateTtir(
+    allocator: std.mem.Allocator,
+    io: std.Io,
+    backend: []const u8,
```
Corendos, line 207: This could be an enum

zml/attention/triton.zig
```diff
@@ -201,7 +201,13 @@ pub const GenerationConfig = union(KernelKind) {
...
+    kernel: []const u8,
```
Corendos, line 208: This could probably also be an enum

zml/attention/triton.zig
Corendos, line 1: I think we could encode the `backend` + `kernel` name into unions of enums.

And we could even have a global type encoding all the possibilities. That would simplify the serialization code.

Something of a superset of the previous `GenerationConfig`
```zig
const GenerationConfg = union(enum) {
    attention: AttentionConfig, // Rename the old GenerationConfig to AttentionConfig
    moe: MoeConfig,
};
```

Here we are essentially declaring a portable schema beween Zig and Python

zml/moe/triton/triton_kernels/__init__.py
Corendos, line 1: I'm not sure it's required to export the symbols from `triton_kernels`

zml/moe/triton/triton_kernels/config.json
Corendos, line 1: 🥲

examples/llm/models/qwen3_5_moe/model.zig
```diff
@@ -0,0 +1,1348 @@
...
+            zml.attention.attention.Metadata.init(.fromBackend(.vanilla, x.dim(.s), self.num_heads)),
+            zml.attention.attention.Parameters.init(.fromBackend(.vanilla)),
+        ).withPartitioning(.{ .q = .replicated, .h = .model, .hd = .replicated }).rename(.{ .q = .s }).merge(.{ .d_out_proj = .{ .h, .hd } });
+
+        const gated_output = attn_output.mul(gate.sigmoid());
+        const projected_output = self.o_proj.forward(gated_output.rename(.{ .d_out_proj = .d }))
+            .rename(.{ .dout = .d }).withPartitioning(.{ .d = .replicated });
+
+        return .{ projected_output, new_kv_cache };
+    }
```
Corendos, line 618: I think this can be simplified, there is a lot of duplication

* Keep the `withPartitioning` where it matters
* `position_ids` is a bit verbose

examples/llm/models/qwen3_5_moe/model.zig
```diff
@@ -0,0 +1,1348 @@
...
+            null,
+            moe_metadata,
+            moe_parameters,
+        ) catch |err| stdx.debug.panic("moe backend failed: {}", .{err});
+
+        const shared_gate = self.shared_expert_gate.forward(x).sigmoid().broad(x.shape());
+        const shared = self.shared_expert.forward(x).mul(shared_gate);
+
+        return moe_output.add(shared);
+    }
```
Corendos, line 747: This can be collapsed no ?

zml/moe.zig
```diff
@@ -0,0 +1,130 @@
...
+    scales_down: ?zml.Tensor,
+    bias_down: ?zml.Tensor,
+    metadata: Metadata,
+    parameters: Parameters,
+) !zml.Tensor {
+    return switch (parameters) {
+        .triton => b: {
+            const triton_metadata = switch (metadata) {
+                .triton => |v| v,
+            };
```
Corendos, line 112: Just add an assert at the top level checking that the two active tags are equal instead of the double switch

zml/moe/triton.zig
```diff
@@ -0,0 +1,816 @@
...
+    else if (num_assignments < num_experts)
+        num_assignments * block_size_m
+    else
+        num_assignments + num_experts * (block_size_m - 1);
+
+    var threaded_io: std.Io.Threaded = .init_single_threaded;
+    threaded_io.allocator = std.heap.c_allocator;
+    defer threaded_io.deinit();
+
+    const io = threaded_io.io();
```
Corendos: Use the `io` exposed by the `CompilationContext`

zml/moe/triton.zig
```diff
@@ -0,0 +1,816 @@
...
+        const naive_num_tokens_post_padded = Tensor.constant(.{ .i32 = @as(i32, @intCast(max_num_tokens_padded)) }).reshape(.{1});
+        break :blk .{ naive_sorted_ids, naive_expert_ids, naive_num_tokens_post_padded };
+    } else try alignBlockSize(std.heap.c_allocator, io, ids, num_experts, block_size_m);
+
+    var hidden_quant = hidden;
+    var a_scale = opts.a1_scale orelse Tensor.scalar(1.0, .f32);
+
+    // If fp8 quantized model, quantize activations to fp8 with per-token-group quantization
+    if (gate_up.dtype() == .f8e4m3fn) {
+        hidden_quant, a_scale = try quantizePerTokenGroupFp8(std.heap.c_allocator, io, hidden, fp8ActivationGroupSize(hidden));
```
Corendos: Use the `allocator` exposed by the `CompilationContext`

zml/moe/triton.zig
```diff
@@ -0,0 +1,816 @@
...
+
+    out.block_size_m = block_size_m.integer;
+    out.block_size_n = block_size_n.integer;
+    out.block_size_k = block_size_k.integer;
+    out.group_size_m = group_size_m.integer;
+    out.num_warps = num_warps.integer;
+    out.num_stages = num_stages.integer;
+
+    return out;
+}
```
Corendos, line 776: Do you think it would be possible to embed the JSON configuration file directly as a constant somewhere instead of loading it everytime we need it ?

zml/moe/triton.zig
```diff
@@ -0,0 +1,816 @@
...
+    pub fn init(opts: InitOptions) Parameters {
+        return .{
+            .num_experts_per_tok = opts.num_experts_per_tok,
+        };
+    }
+};
+
+pub const Metadata = struct {
+    w1_zero_bias: ?Tensor = null,
+    w2_zero_bias: ?Tensor = null,
```
Corendos, line 421: What's the purpose of these tensors ?

Corendos (commented):
Left some comments

Some things to address globally:
* Try to remove the uses of `std.heap.c_allocator` where possible

zml/moe/triton/moe_compile.py
loupicaaa: Moved Fake tensor, contiguous strides and triton dtype into triton/generation/kernel_utils

examples/llm/models/qwen3_5_moe/model.zig
loupicaaa: I updated with the SelfAttn from non moe implementation (cleaner)

zml/moe/triton.zig
loupicaaa: The kernel expect a bias. If there are not for the weights in the model then I give him a 0 tensor created in metadata to avoid paying this potential cost at runtime.

PR #506, opened 2026-04-24
Remove BoundedArray.init 

gwenzek:
When looking at the new Sharding code I was triggered by the high number of "try".

I then realized most of the try where traced back to `axes = try .init(0)` which can never fail.
So I removed BoundedArray.init (aka Andrew mode)

It's generally unsafe because it makes `slice()` return undefined.;
I replaced`try init(0)` by `.empty`
I replaced `try init(n)` by `.fromSlice` or with `.{ .buffer = undefined, .len = n }` to explicitly show the unsafeness

This leads to widespread changes because it means that most of the sharding operations can't fail anymore.

### DIFF
--- a/zml/exe.zig
+++ b/zml/exe.zig
@@ -73,9 +73,10 @@ pub const Exe = struct {
 
     pub const FlatBuffers = struct {
         buffers: []const [*]*pjrt.Buffer,
-        raw_buffers: []const *pjrt.Buffer,
+        raw_buffers: []*pjrt.Buffer,
 
-        num_devices: usize,
+        num_devices: u32,
+        buffer_count: u32,
 
         pub fn init(allocator: std.mem.Allocator, count: usize, num_devices: usize) !FlatBuffers {
             const raw_buffers = try allocator.alloc(*pjrt.Buffer, num_devices * count);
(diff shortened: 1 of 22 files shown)


zml/tensor.zig
```diff
@@ -525,6 +525,7 @@ pub const Tensor = struct {
             return .{ ._state = .init(.{2}, .u64) };
         }
 
+        // TODO(codex): swap the field around to match fromBytes signature: io, platform, sharding, seed
```
Corendos: Maybe I'm weird, but it feels wrong to keep a `TODO(codex)` 🤭

zml/exe.zig
```diff
@@ -73,7 +73,7 @@ pub const Exe = struct {
 
     pub const FlatBuffers = struct {
         buffers: []const [*]*pjrt.Buffer,
-        raw_buffers: []const *pjrt.Buffer,
+        raw_buffers: []*pjrt.Buffer,
```
Corendos: Not sure this change is related, is it part of other changes ?

Corendos (approved):

zml/exe.zig
gwenzek: oopsy

zml/tensor.zig
gwenzek: We need to talk about this one and decide what's the right order, it's very inconsistent across the different functions.

The Zig convention seems to be io first: https://ziggit.dev/t/0-16-0-bikeshed-io-before-alloc-alloc-before-io-not-an-issue/15020/16

(removed from the PR)

PR #512, opened 2026-04-28
examples/llm: cleanups

gwenzek:
I'm doing a pass on the example code we have.

* llama/model.zig, qwen/model.zig: change layer_index from being a tensor to being an integer.
It's clearer this way what is happening. Note that lfm doesn't have the issue because the layer index is passed everywhere.

* tokenizePrompt: the code was doing unnecessary copies of data, I added some helper to tokenizer.encoder to streamline the code by removing the extra copies and allocations.

* `.stdout().writer() -> .stdout().writerStreaming()`  this one is weird but apparently using stdout by position isn't always allowed, in particular when running the exe from within SublimeText.

* runDecode -> show case bake

Considered but dropped

* tokenizer -> add a DecoderWriter to the iree tokenizer because the API perfectly map to the std.Io.Writer API. I feel it would be good to leverage it. It's possible to implement it for other writers too for the sake of uniformity. Not sure we care about the other tokenizers though.


examples/llm/main.zig
```diff
@@ -122,16 +122,21 @@ pub fn main(init: std.process.Init) !void {
     var progress = std.Progress.start(io, .{ .root_name = args.model });
     errdefer progress.end();
 
-    var tokenizer = try loadTokenizer(allocator, io, repo, &progress);
+    var fut_tokenizer = io.async(loadTokenizer, .{ allocator, io, repo, &progress });
+
+    var fut_compiled_model = io.async(models.LoadedModel.compile, .{ &model, allocator, io, platform, backend, shardings, args.seqlen, &progress });
+
+    // Load buffers after the model compilation to be sure to give enough room to the autotune.
```
Finistere: As the comments pointed out, model buffers are loaded after the compilation of the programs because we can't predict how much memory will be consumed during autotuning. So loading everything at once may fail depending on circumstances.

zml/tokenizer/iree/tokenizer.zig
```diff
@@ -463,6 +463,147 @@ pub const Tokenizer = struct {
             }
         }
     };
+
+    pub const DecoderWriter = struct {
```
Finistere: This is a fairly big addition with what limited impact IMHO. For the performance sensitive stuff, we're decoding tokens one by one anyway. And otherwise decodeAlloc from the parent tokenizer is fine. So I don't see the benefit of maintaining this code.

zml/tokenizer/tokenizer.zig
```diff
@@ -40,6 +40,23 @@ pub const Tokenizer = union(Tokenizers) {
...
+                    try e.finalize(writer);
+                },
+                inline else => |*e| {
+                    const encoded: []const u32 = try e.encode(text);
+                    try writer.writeAll(std.mem.sliceAsBytes(encoded));
+                },
+            };
+        }
+
+        pub fn appendTokens(_: *Encoder, writer: *std.Io.Writer, tokens: []const u32) !void {
```
Finistere, line 57: unrelated to the encoder.

examples/llm/main.zig
gwenzek: that's a bit sad ^^' is the autotuner also using xla allocated memory or it's doing it's own cumemalloc on the side ?

zml/tensor.zig
```diff
@@ -521,17 +520,19 @@ pub const Tensor = struct {
         _state: Tensor,
         algorithm: dialects.stablehlo.RngAlgorithm.Type = .DEFAULT,
 
+        pub const Buffer = mem.Bufferized(Rng);
```
Finistere, line 523: Hum interesting, maybe we should use that pattern instead of `zml.Bufferized(Model)`

zml/tokenizer/tokenizer.zig
gwenzek: see how it's used. the point of this code is to make it cleaner to write tokenizer code, see in llama/session.zig:

```

    pub fn tokenizePrompt(self: *const Session, allocator: std.mem.Allocator, prompt: []const u8) ![]const u32 {
        const eot = self.tokenizer.tokenId("<|eot_id|>") orelse return error.NoSuchToken;
        const newline = self.tokenizer.tokenId("\\n") orelse return error.NoSuchToken;

        var tokens = std.Io.Writer.Allocating.initAligned(allocator, .of(u32));
        try tokens.ensureUnusedCapacity(prompt.len);

        const w: *std.Io.Writer = &tokens.writer;
        try encoder.appendTokens(w, &.{ self.config.bos_token_id, start_header });
        try encoder.encode(w, "user");
        try encoder.appendTokens(w, &.{ end_header, newline });
        try encoder.encode(w, prompt);
        try encoder.appendTokens(w, &.{ eot, newline, start_header });
        try encoder.encode(w, "assistant");
        try encoder.appendTokens(w, &.{ end_header, newline });

        return @ptrCast(@alignCast(try tokens.toOwnedSlice()));
    }
```

zml/tokenizer/iree/tokenizer.zig
gwenzek: note this would be in place of the current decoder api.

I think there is some value in providing an API that integrate well into the Zig ecosystem.

(decodeAlloc is typically what I want to avoid)

zml/tokenizer/tokenizer.zig
Finistere: It's just as clear:
```zig
try w.writeAll(&.{ self.config.bos_token_id, start_header });
try encoder.encode(w, "user");
```

examples/llm/main.zig
Finistere: I don't know to be honest

zml/tensor.zig
gwenzek: the pattern already exist for more complex structs like llama model

zml/tokenizer/iree/tokenizer.zig
Finistere: Fair enough

Finistere (approved):

zml/tokenizer/tokenizer.zig
gwenzek: except that doesn't compile, and the pointer cast is a bit ugly. I think it's fine to have a helper method here to make things nicer to read.

zml/tokenizer/iree/tokenizer.zig
gwenzek: I've removed it. It's not really helping examples, if we need it later it's easy enough to bring back.

In general I feel the full zml/tokenizer API should be rework to match more closely what IREE is doing which is great. Then sentencepiece tokenizer could be adapated to match this API, and other tokenizers can be dropped at this point.

examples/llm/main.zig
gwenzek: reverted

zml/tokenizer/tokenizer.zig
gwenzek: kept as agreed IRL

gwenzek:
Thanks all for the discussion !

PR #514, opened 2026-04-28
zml/sharding: pass sharding by pointer and simplify replicated usage

gwenzek:
This PR simplifies the common replicated case while making sharding intent explicit and use sharding by pointer ( sharding struct is quite big: ~600 bytes).

Using pointers for shardings also avoids unnecessary sharding copies and makes it easier to share stable sharding instances across compilation, loading, and runtime paths.

- `Platform` owns the canonical replicated sharding
- `Buffer` APIs now take explicit sharding specs:
    - `.replicated`
    - `.{ .sharded = sharding_ptr }`
- `createTensor` / `maybeCreateTensor `now require explicit partitioning instead of accepting null
- compilation/loading now use pointer-based sharding lists

This makes simple cases like mnist and benchmark much cleaner: callers no longer need to manually construct and thread a replicated sharding just to use the default case, but they still state placement explicitly at the call site.


zml/module.zig
```diff
@@ -44,7 +44,7 @@ fn mlirRegistry(io: std.Io) *mlir.DialectRegistry {
     return mlir_global_registry.?;
 }
 pub const CompilationOptions = struct {
-    shardings: []const Sharding,
+    shardings: []const *const Sharding,
     // If null, will be initialized from the target
     partitioner: ?Partitioning.Partitioner = null,
     // Debugging options
```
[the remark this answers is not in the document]
hugomano: done

zml/io.zig
```diff
@@ -1150,10 +1148,9 @@ pub fn load(
                     defer reader.deinit();
 
                     const shape = reader.tensor.shape;
-                    const select_sharding = sharding_.pickSharding(ctx_.shardings, shape, .explicit_axis_binding);
-                    const sharding = if (select_sharding) |s| s else blk: {
-                        log.debug("No sharding strategy found for tensor {s} with shape {f}, using replicated sharding:\n {f}", .{ reader.tensor.name, shape, ctx_.replicated_sharding });
-                        break :blk ctx_.replicated_sharding;
+                    const sharding = sharding_.pickSharding(ctx_.shardings, shape, .explicit_axis_binding) orelse blk: {
+                        log.debug("No sharding strategy found for tensor {s} with shape {f}, using replicated sharding", .{ reader.tensor.name, shape });
+                        break :blk &ctx_.replicated_sharding;
```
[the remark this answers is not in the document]
hugomano: will be done in another pr

zml/sharding.zig
```diff
@@ -1117,7 +1122,7 @@ pub const Sharding = struct {
     }
 
     pub fn physicalView(self: Sharding) PhysicalView {
```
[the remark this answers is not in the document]
hugomano: done

zml/exe.zig
```diff
@@ -34,17 +34,17 @@ pub const Exe = struct {
...
         var arena = std.heap.ArenaAllocator.init(allocator);
         errdefer arena.deinit();
 
         const input_shapes_copy = try arena.allocator().dupe(Shape, input_shapes);
         const output_shapes_copy = try arena.allocator().dupe(Shape, output_shapes);
 
-        const input_shardings_copy = try arena.allocator().dupe(Sharding, input_shardings);
-        const output_shardings_copy = try arena.allocator().dupe(Sharding, output_shardings);
+        const input_shardings_copy = try arena.allocator().dupe(*const Sharding, input_shardings);
+        const output_shardings_copy = try arena.allocator().dupe(*const Sharding, output_shardings);
```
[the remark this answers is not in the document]
hugomano: done

zml/nn.zig
```diff
@@ -127,10 +127,10 @@ test normalizeL2 {
 
     const input: zml.Tensor = .init(.{ 2, 2 }, .f32);
 
-    var exe = try zml.module.compile(std.testing.allocator, std.testing.io, normalizeL2, .{ input, 1e-12 }, platform, .{ .shardings = &.{zml.testing.replicatedSharding()} });
+    var exe = try zml.module.compile(std.testing.allocator, std.testing.io, normalizeL2, .{ input, 1e-12 }, platform, .{});
```
gwenzek, line 130: I like that we don't have to pass replicated sharding to compile :-)

zml/sharding.zig
```diff
@@ -959,6 +959,8 @@ pub const LogicalMesh = struct {
     axes: Axes,
     intents: Intents,
 
+    pub const replicated = init("replicated", .{ .x = .high_bandwidth });
```
gwenzek: given ZML is creating the replicated sharding for everyone, and also we are comparing sharding based on pointer equality,
we should probably remove the helper function here, to make it harder to create another replicated sharding.

zml/testing.zig
```diff
@@ -332,14 +324,12 @@ pub fn testLayer(
...
-            tensor.* = ctx_.activation_store.createTensor(subkey, null, null);
+            tensor.* = ctx_.activation_store.createTensor(subkey, null, .replicated);
             ctx_.index += 1;
         }
     }.cb, &ctx, &args);
 
-    const replicated_sharding = try zml.sharding.replicatedSharding(platform);
-
-    const exe = try platform.compile(allocator, io, layer, func, args, .{ .shardings = if (shardings.len != 0) shardings else &.{replicated_sharding} });
+    const exe = try platform.compile(allocator, io, layer, func, args, .{ .shardings = if (shardings.len != 0) shardings else &.{} });
```
gwenzek, line 327: `.shardings = shardings`

examples/llm/models/lfm2/inference.zig
```diff
@@ -237,10 +237,9 @@ pub const ComposedKernelExe = struct {
         };
         defer hidden_buf.deinit();
 
-        const replicated_sharding = try zml.sharding.replicatedSharding(args.platform);
-        var conv_cache_index_buf: zml.Buffer = try .scalar(args.io, args.platform, @as(u32, 0), .u32, replicated_sharding);
+        var conv_cache_index_buf: zml.Buffer = try .scalar(args.io, args.platform, @as(u32, 0), .u32, .replicated);
```
gwenzek: note the `@as(u32,...)` isn't required here.

zml/buffer.zig
```diff
@@ -152,22 +163,22 @@ pub const Buffer = struct {
...
     }
 
-    pub fn fromSliceOpts(io: std.Io, platform: *const Platform, slice: Slice, sharding: Sharding, opts: FromOptions) !Buffer {
+    pub fn fromSliceOpts(io: std.Io, platform: *const Platform, slice: Slice, sharding: ShardingSpec, opts: FromOptions) !Buffer {
         return from(io, platform, slice.shape, sharding, std.mem.sliceAsBytes(slice.constData()), opts);
     }
 
     /// Creates a Buffer with a single element.
-    pub fn scalar(io: std.Io, platform: *const Platform, val: anytype, dtype_: DataType, sharding: Sharding) !Buffer {
+    pub fn scalar(io: std.Io, platform: *const Platform, val: anytype, dtype_: DataType, sharding: ShardingSpec) !Buffer {
```
gwenzek: scalar should always be replicated, so you can drop the argument

zml/buffer.zig
```diff
@@ -26,11 +26,22 @@ const log = std.log.scoped(.zml);
 pub const Buffer = struct {
     _platform: *const Platform,
     _shape: Shape,
-    _sharding: Sharding,
+    _sharding: *const Sharding,
     _shards: Shards,
 
     pub const MAX_NUM_SHARDS: u16 = Platform.MAX_NUM_DEVICES;
     pub const Shards = stdx.BoundedArray(*pjrt.Buffer, MAX_NUM_SHARDS);
+    pub const ShardingSpec = union(enum) {
```
gwenzek: I would make this a struct that it has the same size than a pointer. Then we can use it everywhere instead of *const Sharding for the same cost.

```
const ShardingSpec = struct {
   data: ?*const Sharding

  pub const replicated: ShardingSpec = .{ .data = null };
  pub fn sharded(sharding: *const Sharding) {...}
}
```

gwenzek (commented):
Thanks for following up on this @hugomano 

I like the changes and the 'ShardingSpec'

But I'm also wondering if it could make `ShardingSpec` the new `Sharding` and relegate `Sharding` to `Sharding.Data` as an internal type. 
That way there would be only one Sharding object for the full API (compile and buffer manipulation) and we would pass `zml.Sharding` everywhere instead of `*const zml.Sharding`.

Note that this pattern is frequent in C where "cuda_stream" is in fact an alias to `*cuda_stream_data`

The required change would be either:
 - `Sharding.init` takes an allocator to put the `Sharding.Data` on the heap, 
 - use the platform arena to allocate sharding objects, that way we know the user can't mess up, and have shardings live less than their relevant platform.

zml/buffer.zig
Finistere: oh interesting, could just be a `struct` though since there is only one field, no? Definitely feels more consistent to have `.replicated` and `.sharded(...)`

zml/io.zig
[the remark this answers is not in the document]
gwenzek: fixed, by storing replicated sharding inside the platform

zml/exe.zig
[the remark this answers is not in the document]
gwenzek: reverted by storing all sharding object into the platform arena, therefore making them live longer than the executable

hugomano (approved):
Neat!!

zml/io.zig
```diff
@@ -1197,9 +1193,9 @@ fn buildMesh2x2(
     allocator: std.mem.Allocator,
     target: @import("platform.zig").Target,
     devices: []const @import("platform.zig").Device,
-) !sharding_.PhysicalMesh {
+) anyerror!Sharding.PhysicalMesh {
```
Corendos: Is the `anyerror` required here ?

zml/zml.zig
```diff
@@ -39,7 +39,8 @@ pub const Target = platform.Target;
 pub const safetensors = @import("safetensors.zig");
 pub const shape = @import("shape.zig");
 pub const Shape = shape.Shape;
-pub const sharding = @import("sharding.zig");
+pub const Sharding = @import("Sharding.zig");
+pub const sharding = Sharding;
```
Corendos: Not sure about the alias

Corendos (approved):

zml/zml.zig
gwenzek: true, I thought it was going to help, but actually most importer are broken one way or another

zml/io.zig
gwenzek: nope, will remove

PR #517, opened 2026-04-29
zml: replace Python Triton/Mosaic IR generation with native Zig DSLs

elogir:
Drops the Python build-time dependency for Triton and Mosaic TPU kernels. ZML now emits TTIR and Mosaic TPU IR directly from Zig, against the same MLIR context as the rest of the compiler. So no more JAX, Triton at build time and a lighter sandbox.

Split into two commits:

- **`c72f44884` — Add Triton and Mosaic Zig DSL.** Additive: IR builders (`kernels/triton/`, `kernels/mosaic_tpu/`), upstream dialect bindings (`ttir`, `mosaic_tpu`, plus `arith`/`scf`/`math`/`memref`/`vector`/ `affine`/`cf`), and the high-level `zml.kernel.{triton,mosaic_tpu}.Kernel(Config, spec)` declarative form. Comes with author guides in each `README.md`.

- **`cb61e529f` — Remove Python and replace with Zig kernels.** Migrates the two production kernels (`unified_attention` for Triton, `ragged_paged` for Mosaic TPU) and deletes everything Python related in the sandbox and what was generating the IR.


steeve:
rebasing on master

steeve (approved):

PR #521, opened 2026-05-04
workspace: migrate to Bazel 9

dzbarsky:


.bazelrc
```diff
@@ -72,9 +70,11 @@ common          --incompatible_enable_proto_toolchain_resolution
 
 # Allow empty globs since some legacy rules
 common          --incompatible_disallow_empty_glob=false
+# Vendored projects still contain stale layering_check metadata with the Bazel 9 toolchain.
+common          --features=-layering_check
 
 # Use the C linker to link Zig code
-common          --@rules_zig//zig/settings:use_cc_common_link=true
+common          --@rules_zig//zig/settings:use_cc_common_link=false
```
cerisier: Why this ? We sadly need this badly.

dzbarsky: I think it wasn't compatible with the starlarkified rules_cc impl? But I don't recall the details right now, I'll double check in a bit

PR #526, opened 2026-05-07
mlir: add shardy dialect

gwenzek:
Add bindings to shardy dialect,
remove the string generation code from zml.Sharding, and replace it with mlir calls.

### DIFF
--- a/zml/ops.zig
+++ b/zml/ops.zig
@@ -1305,8 +1305,8 @@ fn manualComputationInternal(
                 .results = .{ .flat = global_result_types },
                 .blocks = &.{manual_block},
                 .attributes = &.{
-                    .named(ctx.mlir_ctx, "in_shardings", in_shardings_attr),
-                    .named(ctx.mlir_ctx, "out_shardings", out_shardings_attr),
+                    .named(ctx.mlir_ctx, "in_shardings", in_shardings_attr.asAttr()),
+                    .named(ctx.mlir_ctx, "out_shardings", out_shardings_attr.asAttr()),
                     .named(ctx.mlir_ctx, "manual_axes", manual_axes_attr.asAttr()),
                 },
                 .verify = true,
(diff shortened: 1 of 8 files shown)


hugomano (approved):
Thanks @gwenzek!

zml/ops.zig
```diff
@@ -1308,9 +1305,9 @@ fn manualComputationInternal(
                 .results = .{ .flat = global_result_types },
                 .blocks = &.{manual_block},
                 .attributes = &.{
-                    .named(ctx.mlir_ctx, "in_shardings", in_shardings_attr),
-                    .named(ctx.mlir_ctx, "out_shardings", out_shardings_attr),
-                    .named(ctx.mlir_ctx, "manual_axes", manual_axes_attr),
+                    .named(ctx.mlir_ctx, "in_shardings", in_shardings_attr.asAttr()),
+                    .named(ctx.mlir_ctx, "out_shardings", out_shardings_attr.asAttr()),
+                    .named(ctx.mlir_ctx, "manual_axes", manual_axes_attr.asAttr()),
```
Corendos, line 1310: I wonder if `sdyPerValueShardingAttr` should directly return a `mlir.Attribute` instead. That's kind of what we did for stablehlo because it was not so common to require the typed attribute (and in those rare cases, you can easily `@ptrCast` to the type)

Corendos (approved):
Left a small comment but otherwise LGTM

zml/ops.zig
gwenzek: Shardy dialect is very hierarchical, so I decided to use the explicit types to help with constructing them.

I could change the logic for the top level attribute PerValueSharding but that feels a bit weird.

I can also make it more fluent by using `.attr(in_sharding_attr)` but that doesn't really save much.

hugomano: available here: https://github.com/zml/zml/pull/527

PR #527, opened 2026-05-11
zml/sharding: fix leak and use mlir.Attribute instead of concrete types

hugomano:


gwenzek (approved):
Thamks for the follow up,

I've included a fix for the pointer stability of the sharding objects.

PR #529, opened 2026-05-11
zml/moe : Moe backend triton zig, qwen3_5_moe model

loupicaaa:
### Overview
Context: Modern models utilize a MLP called Mixture of Experts (MoE). In a MoE, the MLP is organized into "experts," where specific sub-tensors are specialized to process particular linguistic patterns or elements. During inference, the hidden states of each token are routed to the most relevant experts, ensuring that matrix multiplications (matmul) are performed only on the active parameters.

Problem: Due to the sparse and dynamic nature of MoE where token assignment is determined at runtime via routing, treating an MoE layer as a standard matmul results in significant performance degradation. Standard kernels are not optimized for the irregular memory access patterns in sparse routing.

Approach: MoE is implemented as a primitive in the ZML operation set. The generic MoE Backend is designed to support various specialized kernels tailored to specific hardware and weight precisions. The primary implementation currently features a Triton MoE backend optimized for NVIDIA Hopper architecture using BF16 and FP8 weights. This implementation is utilized in the Qwen 3.5 MoE model.

### Features
**Generic MoE Backend (zml/moe/moe.zig)**
- Integrate specialized MoE backends (e.g., Triton).
- Select the optimal backend based on hardware and precision.
- Implements essential utilities and structures for registration, parameter loading, and metadata management.
- The forwardMoe() function acts as the entry point to invoke the appropriate backend.

**Triton MoE Backend (zml/moe/triton.zig)**
- Generates TTIR (Triton Intermediate Representation) at compile time.
- Leverages Triton's custom calls exposed via XLA.
Execution Flow:
2. Determine token distribution: use naive assignment if the distribution is sparse; otherwise, sort and align tokens for efficiency.
3. Quantization: quantize activations if using FP8 weights.
4. Fused Kernel: execute the fused MoE kernel for the gate and up-projection layers.
5. Post-Processing: split the output and apply the activation function.
6. Quantization : re-quantize activations for the next stage if required by the FP8 path.
7. Fused Kernel: execute the fused MoE kernel for the down-projection layer.
8. Aggregation: sum the Top-K results for each token.

Qwen 3.5 MoE
- Fully utilizes the Triton MoE backend.
- BF16 weights implementation (also tested with FP8 but not implemented here)

### DIFF
--- a/examples/llm/models/qwen3_5_moe/model.zig
+++ b/examples/llm/models/qwen3_5_moe/model.zig
@@ -908,7 +908,7 @@ pub const GatedDeltaNet = struct {
         RmsNormGated.unloadBuffers(&self.norm);
     }
 
-    fn recurrent_gated_delta_rule(query: zml.Tensor, key: zml.Tensor, value: zml.Tensor, g: zml.Tensor, beta: zml.Tensor, initial_state: ?zml.Tensor) struct { zml.Tensor, zml.Tensor } {
+    fn recurrentGatedDeltaRule(query: zml.Tensor, key: zml.Tensor, value: zml.Tensor, g: zml.Tensor, beta: zml.Tensor, initial_state: ?zml.Tensor) struct { zml.Tensor, zml.Tensor } {
         const scale: f32 = 1.0 / @sqrt(@as(f32, @floatFromInt(query.dim(.khd))));
         const query_norm = zml.nn.normalizeL2(query.rename(.{ .kh = .vh }), 1e-6);
         const key_norm = zml.nn.normalizeL2(key.rename(.{ .kh = .vh }), 1e-6);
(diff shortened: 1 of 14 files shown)


examples/llm/models/qwen3_5_moe/model.zig
```diff
@@ -0,0 +1,1376 @@
...
+        self.in_proj_a.weight.deinit();
+        self.out_proj.weight.deinit();
+        if (self.out_proj_scale) |*scale| scale.deinit();
+        self.conv1d_weight.deinit();
+        self.dt_bias.deinit();
+        self.aLog.deinit();
+        RmsNormGated.unloadBuffers(&self.norm);
+    }
+
+    fn recurrent_gated_delta_rule(query: zml.Tensor, key: zml.Tensor, value: zml.Tensor, g: zml.Tensor, beta: zml.Tensor, initial_state: ?zml.Tensor) struct { zml.Tensor, zml.Tensor } {
```
Corendos: This should be `recurrentGatedDeltaRule`

Corendos (approved):

PR #534, opened 2026-05-12
zml/tensor: add `Tensor.onMemory()`

Corendos:
It is now required with newer plugin to annotate the executable arguments with their location.

Like `Tensor.reuseBuffer()`, you can now use `Tensor.onMemory()` to specify on which memory a tensor is located.


gwenzek (commented):
is memory kind propagated ?

`const y = x.onMemory(.host).addConstant(1)`

Where is y ?

gwenzek (approved):

Corendos:
> Where is y ?

Needs double checking but I think XLA propagates the memory placement, so the computation is done on the host

PR #536, opened 2026-05-12
zml/attention: big rework of paged attention

Corendos:
List of changes:
* Removed `Context`
* Simplified metadata to not contain scratch buffers. They are created uninitialized on the fly instead (thanks to #535)
* Added helper for memory movement (host_pinned -> device) (thanks to #534)
* FA2/FA3 don't read the `max_seqlen_q` from host anymore. This was error prone. Instead rely on a fixed value for now. (It's error prone because it's baked in the command buffer and can result in wrong value if the value changes but the command buffer is not recorded again).


zml/attention/flashattn.zig
```diff
@@ -1292,62 +1286,61 @@ pub const paged_fa3 = struct {
...
+            return .{
+                .decode_offset = .init(.{}, .i32),
+            };
         }
 
         pub fn allocationSize(self: MixedMetadata) usize {
             var allocation_size: usize = 0;
-            allocation_size += self.out_accum.byteSize();
-            allocation_size += self.host_metadata.byteSize();
+            // TODO: Take scratch buffer into account
```
gwenzek: I don't have context on this one. What's scratch buffer ? is the TODO new, or it's writing down a mental TODO that was already there ?

gwenzek (commented):
@Finistere Is this PR helping or  with https://github.com/zml/zml/pull/450 ? 

In anycase @Corendos it will need to be rebased. I'll let you two handle that.

zml/attention/flashattn.zig
Corendos: It's a TODO I forgot to address 🤭

Corendos: Addressed in [683fb2f](https://github.com/zml/zml/pull/536/commits/683fb2fde67b7b869e7703d8b2b211c57034bcbd)[683fb2f](https://github.com/zml/zml/pull/536/commits/683fb2fde67b7b869e7703d8b2b211c57034bcbd)

gwenzek (approved):

PR #541, opened 2026-05-18
zml/module: update sharding to be ptr

r-chong:
Key Changes:
- Update `shardings` to be a pointer

As per Hugo

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -44,7 +44,7 @@ fn mlirRegistry(io: std.Io) *mlir.DialectRegistry {
     return mlir_global_registry.?;
 }
 pub const CompilationOptions = struct {
-    shardings: []const Sharding = &.{},
+    shardings: []const *const Sharding = &.{},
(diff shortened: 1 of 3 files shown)


zml/module.zig
```diff
@@ -44,7 +44,7 @@ fn mlirRegistry(io: std.Io) *mlir.DialectRegistry {
     return mlir_global_registry.?;
 }
 pub const CompilationOptions = struct {
-    shardings: []const Sharding = &.{},
```
gwenzek, line 47: this doesn't seem correct. zml.Sharding is now a pointer itself, I don't think you need a slice of pointers to pointers

gwenzek (commented):
can you share a way to reproduce the issue you're solving, this doesn't seem correct

zml/module.zig
r-chong: Yes I can close the PR and open an issue

PR #549, opened 2026-05-21
zml/Sharding: Rename Placement.Shard to Placement

Finistere:
Keeping `devices_in_canonical_order: []Device` in the `PhysicalMesh` and inside `Sharding.Data` I've changed `physical: PhysicalMesh` to `physical: *const PhysicalMesh` to have the platform own it now  that PhysicalMesh has a `deinit()`.

### DIFF
--- a/zml/Sharding.zig
+++ b/zml/Sharding.zig
@@ -32,6 +32,10 @@ pub fn numPartitionsForLogicalAxis(sharding: Sharding, logical_axis: anytype) i6
     return sharding.data.numPartitionsForLogicalAxis(logical_axis);
 }
 
+pub fn devicesInCanonicalOrder(sharding: Sharding) []const Device {
+    return sharding.data.physical.devices_in_canonical_order;
+}
+
 pub fn format(sharding: Sharding, writer: *std.Io.Writer) std.Io.Writer.Error!void {
     return sharding.data.format(writer);
 }
@@ -103,9 +107,10 @@ pub const Partitioning = struct {
 
     pub fn localShapeForShape(self: Partitioning, shape: Shape) !Shape {
         const sharding = try self.selectSharding(shape);
-        var shards = Placement.shardIterator(sharding, shape);
-        const shard = (try shards.next()) orelse return error.InvalidPhysicalMesh;
-        return shard.shape.withDefaultPartitioning();
+        const ordered_devices = sharding.devicesInCanonicalOrder();
+        if (ordered_devices.len == 0) return error.InvalidPhysicalMesh;
+        const placement_ = try sharding.placement(shape, ordered_devices[0]);
+        return placement_.shape.withDefaultPartitioning();
     }
 
     pub fn numPartitionsForLogicalAxis(self: Partitioning, shape: Shape, logical_axis: anytype) !i64 {
@@ -228,8 +233,6 @@ pub const Device = struct {
         );
     }
 };
-
-pub const Devices = stdx.BoundedArray(Device, Platform.MAX_NUM_DEVICES);
 
 /// Physical axis tags represent hardware interconnect tiers.
 pub const PhysicalAxisTag = enum {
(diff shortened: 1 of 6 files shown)


zml/buffer.zig
```diff
@@ -218,31 +218,31 @@ pub const Buffer = struct {
...
                 },
             };
             const args: pjrt.Client.CreateUninitializedBufferArgs = .{
-                .dims = shard.shape.dims(),
-                .element_type = pjrtx.bufferTypeFromDtype(shard.shape.dtype()),
+                .dims = placement.shape.dims(),
+                .element_type = pjrtx.bufferTypeFromDtype(placement.shape.dtype()),
                 .layout = layout,
-                .dst = .{ .memory = shard.memory(platform, opts.memory).pjrt_memory },
+                .dst = .{ .memory = placement.memory(platform, opts.memory).pjrt_memory },
```
gwenzek, line 247: placement.memory and in particular placement.device seems to be doing too much work.
Why not use `placement.device.pjrt_device` ?

zml/buffer.zig
```diff
@@ -218,31 +218,31 @@ pub const Buffer = struct {
...
-                        shard.shape.dims(),
+                        pjrtx.bufferTypeFromDtype(placement.shape.dtype()),
+                        placement.shape.dims(),
                     );
                     break :blk default_layout.?.toMemoryLayout();
                 },
                 else => .{
                     .tiled = .{
-                        .minor_to_major = constants.minorToMajor(shard.shape.rank()),
+                        .minor_to_major = constants.minorToMajor(placement.shape.rank()),
```
gwenzek: can you move this out of the for loop ?

zml/buffer.zig
```diff
@@ -218,31 +218,31 @@ pub const Buffer = struct {
...
+        for (res._sharding.devicesInCanonicalOrder()) |device| {
+            const placement = try res._sharding.placement(res._shape, device);
             var default_layout: ?pjrt.DefaultMemoryLayout = null;
             const layout: pjrt.MemoryLayout = switch (platform.target) {
                 .tpu => blk: {
                     default_layout = try platform.pjrt_client.defaultMemoryLayout(
                         platform.pjrt_api,
-                        pjrtx.bufferTypeFromDtype(shard.shape.dtype()),
-                        shard.shape.dims(),
+                        pjrtx.bufferTypeFromDtype(placement.shape.dtype()),
```
gwenzek, line 230: dtype is constant through out the for loop, and I suspect `placement.shape.dims()` to also be.
If it's always true, or at least always true on some platform, it should be moved out of the for loop.

zml/Sharding.zig
```diff
@@ -546,22 +579,20 @@ pub const PhysicalMesh = struct {
         }
     }
 
-    pub fn countDevices(self: PhysicalMesh) usize {
+    pub fn countDevices(self: *const PhysicalMesh) usize {
```
gwenzek: what's the purpose of this ?

zml/Sharding.zig
```diff
@@ -546,22 +579,20 @@ pub const PhysicalMesh = struct {
...
         return self.root.countDevices();
     }
 
-    pub fn devices(self: PhysicalMesh) Devices {
-        var list: Devices = .empty;
-        devicesNodeInto(self.root, &list);
-        return list;
+    pub fn devicesInCanonicalOrder(self: *const PhysicalMesh) []const Device {
```
gwenzek: we can remove it.

zml/Sharding.zig
```diff
@@ -610,29 +641,29 @@ pub const PhysicalMesh = struct {
...
 
-    pub fn axis(self: PhysicalMesh, tag: PhysicalAxisTag) i64 {
+    pub fn axis(self: *const PhysicalMesh, tag: PhysicalAxisTag) i64 {
         if (self.axisInfo(tag)) |info| return info.size;
 
         return 1;
     }
 
-    pub fn axisOrder(self: PhysicalMesh) AxisTraversal.Order {
+    pub fn axisOrder(self: *const PhysicalMesh) AxisTraversal.Order {
```
gwenzek: passing by pointer is a bit pointless in this kind of function.
normally LLVM does a better job when you pass stuff by value, because it means they can't get aliased.
And in any case it should be inlined.

zml/Sharding.zig
```diff
@@ -1184,7 +1215,7 @@ pub const PhysicalView = struct {
 
 pub const Data = struct {
     name: []const u8,
-    physical: PhysicalMesh,
```
gwenzek, line 1189: the idea was that Data was storing everything Sharding related.
Yes it's a huge struct, that's why we don't pass it by value anymore.
Why do you need a pointer indirection here ?

gwenzek (commented):
like the direction, thanks.

I'm wondering if we should go one step further: drop the notion of `Sharding.Device` and use SoA layout.
The sharding object contains two lists one with the id of devices, and another one with their coords. Either we use `*pjrt.Device` as the id of the device, or we use the offset into `platform.devices`

gwenzek:
not sure about the changes to pass by pointer everywhere. Normally Zig / LLVM are pretty good with value semantics on arguments. 

You'll need an "optimization remarks" or profile report to convince me otherwise

https://ziglang.org/documentation/master/#Pass-by-value-Parameters

zml/buffer.zig
Finistere: Will take a look at this in a next PR when thinking about the devices, it's not clear enough in my head right now.

hugomano (approved):

zml/Sharding.zig
gwenzek: it's because each PhysicalMesh is unique per platform so we keep a central copy of it instead of copying it several time (it's big)

PR #556, opened 2026-05-27
zml: make Buffer.from faster

gwenzek:
`Buffer.from` is in the hot path everytime we want to send smth to the GPU.
Sometimes we can overlap it with GPU compute but not always, and sometimes it's quite complicated to do so.

I've rewrote Sharding.Placement to be device agnostic instead of device specific.
This allow to reuse most of the logic for multiple devices.

According to LLVM MCA (with script included here), we got from:

```
Iterations:        100
Instructions:      324200
Total Cycles:      175191
Total uOps:        369700
```

to 

```
    Iterations:        100
    Instructions:      162900
    Total Cycles:      85585
    Total uOps:        182600
```

So about twice as fast for this part of the code. I would need some help to confirm the numbers on TPU.

(Once the PR is approved, I'll remove the mca script as well as the build.zig)

### DIFF
--- a/zml/Sharding.zig
+++ b/zml/Sharding.zig
@@ -1,7 +1,6 @@
 //! Explain how to split a buffer across different devices.
 const std = @import("std");
 
-const c = @import("c");
 const dialects = @import("mlir/dialects");
 const mlir = @import("mlir");
 const pjrt = @import("pjrt");
@@ -217,17 +216,20 @@ pub const Device = struct {
     id: usize,
 
     /// Coordinates in the physical mesh
-    coords: stdx.BoundedArray(u8, MAX_MESH_RANK) = .empty,
+    coords: Coords,
 
-    pub fn coordsSlice(self: *const Device) ?[]const u8 {
-        if (self.coords.len == 0) return null;
-        return self.coords.constSlice();
+    const Coords = [MAX_MESH_RANK]u8;
+    // Placeholder coords, that would most likely trigger panic if used like that.
+    const undefined_coords: Coords = @splat(0xff);
+
+    fn isFullInitialized(device: Device) bool {
+        return @as(u32, @bitCast(device.coords)) != @as(u32, @bitCast(undefined_coords));
     }
 
-    pub fn format(self: Device, writer: *std.Io.Writer) !void {
+    pub fn format(device: Device, writer: *std.Io.Writer) !void {
         try writer.print(
             "Device(id={d} coords={any})",
-            .{ self.id, self.coords.constSlice() },
+            .{ device.id, device.coords },
         );
     }
 };
@@ -297,7 +299,7 @@ pub const PhysicalNode = union(enum) {
         return .{
             .leaf = .{
                 .id = @intCast(device_.id()),
-                .coords = .empty,
+                .coords = Device.undefined_coords,
             },
         };
     }
@@ -436,7 +438,7 @@ pub const PhysicalMesh = struct {
         try validateGeometry(root);
 
         var owned_root = root;
-        var path: [MAX_MESH_RANK]u8 = @splat(0);
+        var path: Device.Coords = @splat(0);
         assignCoords(&owned_root, &path, 0);
 
         var mesh: PhysicalMesh = .{
@@ -446,6 +448,10 @@ pub const PhysicalMesh = struct {
             .devices_in_canonical_order = undefined,
         };
         try mesh.populateDevicesInCanonicalOrder(allocator);
+        for (0.., mesh.devices_in_canonical_order) |i, device| {
+            std.debug.assert(device.id == i);
+            std.debug.assert(device.isFullInitialized());
+        }
         return mesh;
     }
 
@@ -458,10 +464,8 @@ pub const PhysicalMesh = struct {
         const Order = struct {
             mesh: *const PhysicalMesh,
             fn lessThan(ctx: @This(), a: Device, b: Device) bool {
-                const ca = a.coordsSlice() orelse unreachable;
-                const cb = b.coordsSlice() orelse unreachable;
-                const ia = ctx.mesh.linearIndexFromCoords(ca);
-                const ib = ctx.mesh.linearIndexFromCoords(cb);
+                const ia = ctx.mesh.linearIndexFromCoords(a.coords);
+                const ib = ctx.mesh.linearIndexFromCoords(b.coords);
                 return ia < ib;
             }
         };
@@ -558,13 +562,13 @@ pub const PhysicalMesh = struct {
     /// coords:
     ///   (0,0) (0,1)
     ///   (1,0) (1,1)
-    fn assignCoords(node: *PhysicalNode, path: *[MAX_MESH_RANK]u8, depth: usize) void {
+    fn assignCoords(node: *PhysicalNode, path: *Device.Coords, depth: usize) void {
         switch (node.*) {
             .leaf => |*d| {
                 // Coords can already been set by caller
-                if (d.coords.len > 0) return;
+                if (d.isFullInitialized()) return;
 
-                d.coords = .init(path[0..depth]);
+                d.coords = path.*;
             },
             .branch => |*b| {
                 for (b.children, 0..) |*child, i| {
@@ -594,7 +598,7 @@ pub const PhysicalMesh = struct {
     ///
     /// sizes: S0..Sk, coords: C0..Ck
     /// linear = (((C0 * S1) + C1) * S2 + C2) ...
-    pub fn linearIndexFromCoords(self: PhysicalMesh, coords: []const u8) usize {
+    pub fn linearIndexFromCoords(self: PhysicalMesh, coords: Device.Coords) usize {
         const order = self.axisOrder();
         var idx: usize = 0;
         for (order.constSlice()) |tag| {
@@ -706,7 +710,7 @@ pub const PhysicalMesh = struct {
     const CoordsTopology = struct {
         const CoordPlacement = struct {
             device_id: usize,
-            coords: stdx.BoundedArray(u8, MAX_MESH_RANK),
+            coords: Device.Coords,
         };
 
         const CoordLayout = struct {
@@ -722,22 +726,20 @@ pub const PhysicalMesh = struct {
         fn parseDevice(device: PlatformDevice) !CoordPlacement {
             const api = device.platform.pjrt_api;
             const coords_attr = device.pjrt_desc.attribute(api, "coords") orelse return error.MissingDeviceCoords;
-            var coords: stdx.BoundedArray(u8, MAX_MESH_RANK) = .empty;
-            switch (coords_attr) {
+            const coords: Device.Coords = coords: switch (coords_attr) {
                 .int64list => |values| {
+                    var coords: Device.Coords = @splat(0);
                     if (values.len == 0 or values.len > MAX_MESH_RANK) return error.InvalidDeviceCoords;
-                    for (values) |value| {
+                    for (coords[0..values.len], values) |*coord, value| {
                         if (value < 0) return error.InvalidDeviceCoords;
-                        coords.appendAssumeCapacity(@intCast(value));
+                        coord.* = @intCast(value);
                     }
+                    break :coords coords;
                 },
                 else => return error.InvalidDeviceCoords,
-            }
-
-            return .{
-                .device_id = device.id(),
-                .coords = coords,
             };
+
+            return .{ .device_id = device.id(), .coords = coords };
         }
 
         pub fn collect(allocator: std.mem.Allocator, platform_devices: []const PlatformDevice) ![]CoordPlacement {
@@ -763,7 +765,7 @@ pub const PhysicalMesh = struct {
             for (0..rank) |ax_i| {
                 var max_coord: usize = 0;
                 for (placements) |coord_placement| {
-                    max_coord = @max(max_coord, coord_placement.coords.constSlice()[ax_i]);
+                    max_coord = @max(max_coord, coord_placement.coords[ax_i]);
                 }
                 axis_sizes[ax_i] = max_coord + 1;
             }
@@ -774,7 +776,7 @@ pub const PhysicalMesh = struct {
             };
         }
 
-        fn linearIndex(coords: []const u8, rank: usize, axis_sizes: []const usize) usize {
+        fn linearIndex(coords: Device.Coords, rank: usize, axis_sizes: []const usize) usize {
             var linear: usize = 0;
             var stride: usize = 1;
             var ax = rank;
@@ -791,7 +793,7 @@ pub const PhysicalMesh = struct {
             for (placements, indexed) |coord_placement, *out| {
                 out.* = .{
                     .placement = coord_placement,
-                    .linear = linearIndex(coord_placement.coords.constSlice(), rank, axis_sizes),
+                    .linear = linearIndex(coord_placement.coords, rank, axis_sizes),
                 };
             }
 
@@ -809,13 +811,8 @@ pub const PhysicalMesh = struct {
             return indexed;
         }
 
-        pub fn leaf(coord_placement: *const CoordPlacement, coords_slice: []const u8) !PhysicalNode {
-            return .{
-                .leaf = .{
-                    .id = coord_placement.device_id,
-                    .coords = try .fromSlice(coords_slice),
-                },
-            };
+        pub fn leaf(coord_placement: *const CoordPlacement, coords: Device.Coords) !PhysicalNode {
+            return .{ .leaf = .{ .id = coord_placement.device_id, .coords = coords } };
         }
 
         pub fn axisStrides(axis_sizes: []const usize) ![MAX_MESH_RANK]usize {
@@ -843,7 +840,7 @@ pub const PhysicalMesh = struct {
         ) !PhysicalNode {
             if (depth == axis_sizes.len) {
                 const coord_placement = &indexed[base_offset].placement;
-                return leaf(coord_placement, coord_placement.coords.constSlice()[0..axis_sizes.len]);
+                return leaf(coord_placement, coord_placement.coords);
             }
 
             const children = try allocator.alloc(PhysicalNode, axis_sizes[depth]);
@@ -913,8 +910,8 @@ pub const PhysicalMesh = struct {
 
         const nodes = try allocator.alloc(PhysicalNode, platform_devices.len);
         for (nodes, indexed, 0..) |*n, entry, i| {
-            const linear_coord: [1]u8 = .{@intCast(i)};
-            n.* = try CoordsTopology.leaf(&entry.placement, linear_coord[0..]);
+            const linear_coord: Device.Coords = .{ @intCast(i), 0, 0, 0 };
+            n.* = try CoordsTopology.leaf(&entry.placement, linear_coord);
         }
 
         return .{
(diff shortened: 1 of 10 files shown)


zml/buffer.zig
```diff
@@ -106,38 +115,26 @@ pub const Buffer = struct {
...
-                    },
-                },
-            };
+        const placement = try res._sharding.placement(sh);
+        const shard_dims: []const i64 = placement.shape.dims();
+        const layout = platform.defaultMemoryLayout(sh);
+
+        for (platform.physical_mesh.devices_in_canonical_order) |device| {
+            // TODO: detect this failure earlier and out of the for loop.
+            const memory = platform.devices[device.id].memory(opts.memory) orelse @panic("Device doesn't have this memory");
```
gwenzek: It's not clear how to best handle unsupported memory.
Currently Buffer.from crashes when you ask a memory not supported by the platform.

Maybe we should handle that when creating the platform, eg always redirecting host_unpinned to host_pinned.

zml/platform.zig
```diff
@@ -12,21 +13,11 @@ const pjrtx = @import("pjrtx.zig");
...
-    var struct_field_types: [tag_count]type = @splat(T);
-    const is_optional = @typeInfo(T) == .optional;
-    const default_value_ptr: ?*const anyopaque = if (is_optional) @ptrCast(&@as(T, null)) else null;
-    var struct_field_attrs: [tag_count]std.builtin.Type.StructField.Attributes = @splat(.{ .default_value_ptr = default_value_ptr });
-    inline for (@typeInfo(E).@"enum".fields, 0..) |f, i| struct_field_names[i] = f.name;
-    return @Struct(.auto, null, &struct_field_names, &struct_field_types, &struct_field_attrs);
-}
-
-var api_map: StaticPlatformMap(Target, ?*const pjrt.Api) = .{};
+var api_map: std.enums.EnumArray(Target, ?*const pjrt.Api) = .initFill(null);
```
gwenzek, line 20: cosmetic change, but I was wondering why we had a std.enum fork here.

zml/platform.zig
```diff
@@ -128,6 +119,7 @@ pub const Device = struct {
     pjrt_device: *const pjrt.Device,
     pjrt_desc: *const pjrt.DeviceDescription,
     addressable_memories: []*const Memory,
+    memory_by_kind: std.EnumArray(Memory.Kind, ?*const Memory),
```
gwenzek, line 122: cache to avoid loops in the hot path

zml/shape.zig
```diff
@@ -319,8 +317,10 @@ pub const Shape = struct {
         };
     }
 
-    inline fn ensureAttributesAreSync(self: Shape) void {
-        stdx.debug.assert(self._dims.len == self._tags.len and self._dims.len == self._partitioning.len, "Tags, dims and partitioning have diverged! dims={d} tags={d} partitioning={d}", .{ self._dims.len, self._tags.len, self._partitioning.len });
+    fn ensureAttributesAreSync(self: Shape) void {
+        if (builtin.mode == .Debug) {
+            stdx.debug.assert(self._dims.len == self._tags.len and self._dims.len == self._partitioning.len, "Tags, dims and partitioning have diverged! dims={d} tags={d} partitioning={d}", .{ self._dims.len, self._tags.len, self._partitioning.len });
```
gwenzek, line 322: this check is not very useful. If someone manipulate internal object they must know what they are doing.

zml/Sharding.zig
```diff
@@ -1,24 +1,25 @@
...
 const Shape = @import("shape.zig").Shape;
 const Slice = @import("slice.zig").Slice;
 const Target = @import("platform.zig").Target;
+const builtin = @import("builtin");
 
 const Sharding = @This();
 
 data: *const Data,
 
+pub const MAX_MESH_RANK = 4;
```
gwenzek, line 21: this means Mesh is max 4D. I haven't seen more than 3D meshes for now.
this mean the tye coord fits on a u32.

zml/slice.zig
```diff
@@ -18,10 +18,8 @@ pub fn isBytes(comptime T: type) bool {
 }
 
 pub const Slice = struct {
-    inner_data: union(enum) {
-        mutable: []u8,
-        immutable: []const u8,
-    },
+    bytes: []const u8,
+    mutable: bool,
```
gwenzek, line 22: this change was for convenience.
I found it hard to write code that access the bytes directly because of the wrapper union.

But this is not less/more safe.
Because the exposed member is []const u8, it's fine to treat a mutable Slice as a Const one.
When you do `.items` to get a mutable slice only then you check the tag, where as before you'd read the tag on every access

zml/platform.zig
```diff
@@ -136,20 +128,49 @@ pub const Device = struct {
...
 
+        // Cache memory lookups since they are expensive
+        const default_memory: *const Memory = resolveDefaultMemory(pjrt_device_, platform, addressable_memories);
+        const memory_by_kind: std.EnumArray(Memory.Kind, ?*const Memory) = .init(.{
+            .default = default_memory,
+            .device = resolveMemory(addressable_memories, .device),
+            .host_pinned = resolveMemory(addressable_memories, .host_pinned),
+            .host_unpinned = resolveMemory(addressable_memories, .host_unpinned),
+        });
+
```
gwenzek, line 139: I don't really like this. This is effectively replicate current behavior with caching.

But it means if someone call `x.from(.host_unpinned)` on the 5090 it will crash at runtime.
But f they do that on .neuron platform that doesn't have `.host_unpinned` it's translated to the default memory. This behavior is encoded in `Memory.isOfKind`

I think I'd be in favor to resolve potential fallbacks here see eg https://github.com/zml/zml/commit/999927839d56c684bedf71e4dd1f47c33df4c74c#diff-f5d2d55b86d9ef1831d889cacdb9dff94f601f9bfebc20b1d3e5ac4cbb66067b

zml/buffer.zig
```diff
@@ -218,33 +217,21 @@ pub const Buffer = struct {
...
-                },
-                else => .{
-                    .tiled = .{
-                        .minor_to_major = minor_to_major,
-                        .tile_dims = &.{},
-                        .tile_dims_sizes = &.{},
-                    },
-                },
-            };
+        stdx.debug.assert(platform.devices[0].memory(opts.memory) != null, "Device doesn't have {} memory", .{opts.memory});
```
gwenzek, line 218: It's not clear how to best handle unsupported memory.
Currently Buffer.from crashes when you ask a memory not supported by the platform.

Maybe we should handle that when creating the platform, eg always redirecting host_unpinned to host_pinned.

neudinger: The problem is that some platform does not have the extention but we have to support it (like tenstorent)
So the assert is not a good option.

zml/platform.zig
```diff
@@ -567,6 +601,24 @@ pub const Platform = struct {
         }
         unreachable;
     }
+
+    pub fn defaultMemoryLayout(platform: *const Platform, dims: []const i64, dtype: zml.DataType) pjrt.MemoryLayout {
+        return switch (platform.target) {
+            .cuda, .rocm, .tpu, .neuron, .oneapi => {
```
gwenzek: @neudinger this is the new defaultMemoryLayout logic.
Main change is that I don't force materializing the shape object, which I don't have in my updated code.

I also bypass the logic on CPU, because this is always the returned layout.
Ultimately I'm not sure I like the amount of hoops we have to jump to get a memory layout.

zml/Sharding.zig
```diff
@@ -449,6 +449,10 @@ pub const PhysicalMesh = struct {
             .devices_in_canonical_order = undefined,
         };
         try mesh.populateDevicesInCanonicalOrder(allocator);
+        for (0.., mesh.devices_in_canonical_order) |i, device| {
```
gwenzek: this seems to be an implicit assumption of the sharding code, so I enforce it here.

examples/llm/BUILD.bazel
```diff
@@ -11,6 +11,7 @@ zig_binary(
     main = "main.zig",
     visibility = ["//visibility:public"],
     deps = ["//zml", "@linenoise"],
+    zigopts = ["-femit-asm=/tmp/zml/llm.asm"],
```
Corendos: This was part of the perf investigation, we can remove it I guess ?

zml/tensor.zig
```diff
@@ -281,6 +281,46 @@ pub const Tensor = struct {
...
+            }
+        };
+
+        const exe = try zml.module.compile(std.testing.allocator, std.testing.io, Local.memcopy, .{ left, right }, platform, .{});
+        defer exe.deinit();
+
+        var left_d = try zml.Buffer.fromBytes(io, platform, left.shape(), .replicated, @ptrCast(&inputs));
+        defer left_d.deinit();
+
+        var right_d = try zml.Buffer.uninitialized(io, platform, left.shape(), .replicated, .{});
```
gwenzek, line 305: the original idea was having a test for Buffer.uninitialized but I wasn't sure what to test. I thought of this feature which could use some test coverage.

zml/buffer.zig
```diff
@@ -96,6 +104,7 @@ pub const Buffer = struct {
         data_: []const u8,
         opts: FromOptions,
     ) !Buffer {
+        asm volatile ("# LLVM-MCA-BEGIN zml_Buffer_from" ::: .{ .memory = true });
```
Corendos: Ditto

zml/Sharding.zig
```diff
@@ -209,27 +209,28 @@ fn shapeHasAxisPartition(shape: Shape) bool {
...
+    const Coords = [MAX_MESH_RANK]u8;
+    // Placeholder coords, that would most likely trigger panic if used like that.
+    // Those are used when creating a mesh, before calling assignCoords
+    // Private so it's kept as implementation detail
+    const undefined_coords: Coords = @splat(0xff);
 
-    pub fn coordsSlice(self: *const Device) ?[]const usize {
-        if (self.coords.len == 0) return null;
-        return self.coords.constSlice();
+    fn asValidCoords(device: Device) bool {
```
Corendos:
```suggestion
    fn hasValidCoords(device: Device) bool {
```

zml/Sharding.zig
```diff
@@ -209,27 +209,28 @@ fn shapeHasAxisPartition(shape: Shape) bool {
...
-    coords: stdx.BoundedArray(usize, Shape.MAX_RANK) = .empty,
+    coords: Coords,
 
-    /// Placeholder
-    pjrt_device: ?*const pjrt.Device = null,
+    const Coords = [MAX_MESH_RANK]u8;
+    // Placeholder coords, that would most likely trigger panic if used like that.
+    // Those are used when creating a mesh, before calling assignCoords
+    // Private so it's kept as implementation detail
+    const undefined_coords: Coords = @splat(0xff);
```
Corendos, line 224: Just a thought but I remember Andrew advocating for non-exhaustive enums for that usecase. What do you think ?

zml/platform.zig
hugomano: we should ask @steeve on this

hugomano: maybe a Zig 0.16 temp bug

zml/buffer.zig
```diff
@@ -150,6 +147,7 @@ pub const Buffer = struct {
             try res.await(io);
         }
 
+        asm volatile ("# LLVM-MCA-END zml_Buffer_from" ::: .{ .memory = true });
```
Corendos: Ditto

zml/platform.zig
```diff
@@ -567,6 +601,24 @@ pub const Platform = struct {
         }
         unreachable;
     }
+
+    pub fn defaultMemoryLayout(platform: *const Platform, dims: []const i64, dtype: zml.DataType) pjrt.MemoryLayout {
```
hugomano: Will trigger dangling pointer from what I see

build.zig
Corendos, line 1: This and the `.zon` files should be part of another PR I think

gwenzek: yes, I acknowledged in my PR description. It's just important for my local dev so I only remove it at the end

zml/Sharding.zig
gwenzek: I'm not sure how I can use an enum for a [4]u8 ?

zml/platform.zig
gwenzek: thanks I didn't realized that MemoryLayout is pointing into DefaultMemoryLayout.
Feels very weird. I feel there should be a better way of doing this. I put `inline` as a stopgap

gwenzek:
I fixed the crash, thanks @hugomano and removed the local files.

Any opinion on https://github.com/zml/zml/pull/556#discussion_r3316006843 ?

zml/buffer.zig
```diff
@@ -106,38 +114,26 @@ pub const Buffer = struct {
...
-                        .tile_dims_sizes = &.{},
-                    },
-                },
-            };
+        const placement = try res._sharding.placement(sh);
+        const shard_dims: []const i64 = placement.shape.dims();
+        const layout = platform.defaultMemoryLayout(shard_dims, sh.dtype());
+
+        for (platform.physical_mesh.devices_in_canonical_order) |device| {
+            const memory = platform.devices[device.id].memory(opts.memory).?;
```
hugomano, line 126: this is not stable, for some platforms it's the local hardware id to use, for sparse ids we will have out of bound

hugomano: It's also still used in zml/io.zig

zml/Sharding.zig
hugomano: no pjrt_device right there . The CI failed too. https://github.com/zml/zml/actions/runs/26582869314/job/78320533000?pr=556

Also for TPU and Neuron I think the coords will fail cause of tank 4. 


eg platform.fmtVerbose for TPU:
```
platform: tpu
version: 0.103
extensions:
├─ unknown
├─ unknown
├─ unknown
├─ unknown
├─ unknown
├─ unknown
├─ unknown
├─ unknown
├─ unknown
├─ phase_compile
├─ unknown
├─ unknown
├─ unknown
├─ memory_descriptions
├─ layouts
├─ raw_buffer
└─ profiler
plugin attributes:
├─ xla_version: 2
├─ stablehlo_current_version: 1.16.0
└─ stablehlo_minimum_version: 0.9.0
devices:
├─ TPU v6 lite (TPU_0(process=0,(0,0,0,0)))
│  ├─ coords: { 0, 0, 0 }
│  ├─ core_on_chip: 0
│  ├─ device_memory_bytes_limit: 33550235648
│  ├─ num_cores: 1
│  └─ memories:
│     ├─ memory: TpuHbmMemory(id=0, process_index=0, client=tpu)
│     ├─ memory: PinnedHostMemory(id=1, process_index=0, client=tpu)
│     └─ memory: UnpinnedHostMemorySpace(id=2, process_index=0, client=tpu)
├─ TPU v6 lite (TPU_1(process=0,(1,0,0,0)))
│  ├─ coords: { 1, 0, 0 }
│  ├─ core_on_chip: 0
│  ├─ device_memory_bytes_limit: 33550235648
│  ├─ num_cores: 1
│  └─ memories:
│     ├─ memory: TpuHbmMemory(id=3, process_index=0, client=tpu)
│     ├─ memory: PinnedHostMemory(id=4, process_index=0, client=tpu)
│     └─ memory: UnpinnedHostMemorySpace(id=5, process_index=0, client=tpu)
├─ TPU v6 lite (TPU_2(process=0,(0,1,0,0)))
│  ├─ coords: { 0, 1, 0 }
│  ├─ core_on_chip: 0
│  ├─ device_memory_bytes_limit: 33550235648
│  ├─ num_cores: 1
│  └─ memories:
│     ├─ memory: TpuHbmMemory(id=6, process_index=0, client=tpu)
│     ├─ memory: PinnedHostMemory(id=7, process_index=0, client=tpu)
│     └─ memory: UnpinnedHostMemorySpace(id=8, process_index=0, client=tpu)
└─ TPU v6 lite (TPU_3(process=0,(1,1,0,0)))
   ├─ coords: { 1, 1, 0 }
   ├─ core_on_chip: 0
   ├─ device_memory_bytes_limit: 33550235648
   ├─ num_cores: 1
   └─ memories:
      ├─ memory: TpuHbmMemory(id=9, process_index=0, client=tpu)
      ├─ memory: PinnedHostMemory(id=10, process_index=0, client=tpu)
      └─ memory: UnpinnedHostMemorySpace(id=11, process_index=0, client=tpu)
```

and for Neuron:

```
platform: neuron
version: 0.79
extensions:
├─ profiler
└─ layouts
plugin attributes:
├─ stablehlo_current_version: 1.11.0
└─ stablehlo_minimum_version: 1.11.0
devices:
├─ NC_v2 (NeuronCore(id=0, process_index=0, local_id=0))
│  ├─ slice_index: 0
│  └─ memories:
│     ├─ memory: NeuronMemory(id=0, kind=device)
│     └─ memory: NeuronMemory(id=1, kind=pinned_host)
└─ NC_v2 (NeuronCore(id=1, process_index=0, local_id=1))
   ├─ slice_index: 0
   └─ memories:
      ├─ memory: NeuronMemory(id=0, kind=device)
      └─ memory: NeuronMemory(id=1, kind=pinned_host)
```

you see that memory id are the same for both devices. actually inf2.8xlarge has a semantic of 1 device with two Neuron Cores but act as 2 devices in the PjRT world.

zml/buffer.zig
```diff
@@ -76,6 +77,10 @@ pub const Buffer = struct {
         return self._shape;
     }
 
+    pub fn numDevices(self: Buffer) u32 {
+        return @intCast(self._shards.len);
+    }
```
hugomano: It’s both true and not true. As of today, yes but a buffer should not be directly aware of that. If we start doing replicated execution or program parallelism, we’ll run into problems. The idea is that sharding abstracts that away.

That was the whole idea behind not having device.id or device.* references in the codebase outside of zml.sharding when I introduced it cause a device is a primitive for model compilation / execution. A buffer is bound to a memory space nothing else from my perspective and that's the whole story about the iterations on PJRT_Memory API cardinalities the last year from my understanding. Eg: Metal which is ongoing and GB200 are unified memories with 1 device for metal and X devices for GB200. Also maybe a primitive is missing.

gwenzek: I can remove it, but I don't really see the point since you can self._shards.len anyway. This isn't supposed to be the number of devices of the platform, but the number of devices that support this buffer. Maybe call it numShards ?

zml/buffer.zig
gwenzek: My point is that `device.id` should be an offset into `platform.devices`. 
This doesn't have to map to whatever other ID can be used to address the device. Think of it as a `*Device`

gwenzek: see eg https://github.com/zml/zml/pull/562 where I totally replaced Sharding.Device by a *Device

build.zig
gwenzek: done

zml/buffer.zig
gwenzek: I've added some cleanup code in PhysicalMesh.neuron that correctly restore the ids at the end as offset into platform.devices. There is probably a better way to do this, but this ensure that I'm not breaking any of the logic: your code run with your ids, my code run with my ids.

zml/Sharding.zig
gwenzek: I don't think I'm using the assumption one device = one memory. I'm just storing the result of device.getMemory(kind) for each kind (since there are just 3 kinds). It's allowed for a memory to be addressed by two devices.

zml/buffer.zig
gwenzek: done

zml/buffer.zig
hugomano: numShards seems better yes

hugomano (approved):

gwenzek:
Thanks for testing on Neuron !

PR #558, opened 2026-05-27
zml/buffer: use default pjrt memory layout

neudinger:
Alway use default memory layout given by the pjrt plugin.

remove platform switch case evaluation (improve performances in some cases).

Work on: nvidia, intel, rocm, tpu, neuron.


gwenzek:
My point wasn't about the switch but having to fetch the layout independently for all devices.
See https://github.com/zml/zml/pull/556

PR #555, opened 2026-05-27
examples/llm: split Llama and Qwen 3.5 with LFM cleanup

hugomano:
This PR splits Llama in `examples/llm` into separate `embed_tokens`, `layer`, and `lm_head` kernels for both prefill and decode, and brings LFM2 onto the same composed execution path. And done same work for Qwen 3.5 .

It removes the old LFM2 `--single` mode, reworks Llama KV-cache access to use explicit layer indices, and reuses more decode-time state through a persistent runner and prebuilt token index buffers. 

It also tightens sharding/partitioning in LFM2 for multi-device runs, adds clearer per-phase compile logging/program names, removes the unused `--single` CLI flag.

**Llama 3.1 8B on 2x5090**

```
$ CUDA_VISIBLE_DEVICES=0,1 bazel run --config=release --//platforms:cuda=true --//platforms:cpu=false //examples/llm -- --model=file:///var/models/meta-llama/Llama-3.1-8B-Instruct/ --prompt="Tell me a story so I sleep well. I love fantasy settings and weird machines." --seqlen=1024

INFO: Running command line: bazel-bin/examples/llm/llm <args omitted>
warning(zml/io/vfs/hf): No Hugging Face authentication token found in environment or home config; proceeding without authentication.
warning(zml/io/vfs/s3): S3 initialized without full credentials (AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY). Access will be unauthenticated. Using default endpoint: https://s3.amazonaws.com (set AWS_ENDPOINT_URL or AWS_ENDPOINT_URL_S3). Using default region: us-east-1 (set AWS_REGION or AWS_DEFAULT_REGION).
warning(zml/platforms/cuda): Detected NVIDIA GPU that requires CUDA compatibility libraries.
info(zml/platforms/cuda): Loaded CUDA compatibility libraries.
info(pjrt): Loading: libpjrt_cuda.so...
info(pjrt): Loaded: libpjrt_cuda.so
info(llm): Selected backend: .cuda_fa2
info(llm): Resolving model repository..
info(llm): Initializing model..
info(llm): Detected model type: .llama
info(llm): Loaded tokenizer [184.022ms]
info(llama): Compiled prefill embed_tokens [326.51ms]
info(llama): Compiled prefill transformer layer [2.212s]
info(llama): Compiled prefill lm_head [2.219s]
info(llama): Compiled decode embed_tokens [39.536ms]
info(llama): Compiled decode transformer layer [726.752ms]
[...]

PR #560, opened 2026-05-28
examples/llm: remove useless sharding constraints

hugomano:
Perf iso to https://github.com/zml/zml/pull/555


gwenzek (approved):
nice much cleaner

PR #576, opened 2026-06-15
zml/io: expose DMA chunk windows to readers from DirectMemoryWriter

hugomano:
DirectMemoryWriter used to expose the next shard slice as its writable window. For VFS backends like HF this often meant the reader saw tiny buffers and issued tiny reads, even when the next tensor bytes were contiguous enough to load as a larger chunk.

This PR makes the direct writer expose the active primary shard DMA buffer as the public std.Io.Writer buffer. Instead of asking the reader for one small axis slice at a time for a device writer, we let std.Io fill a larger DMA window directly and only rotate that window when a dispatch run, DMA chunk, or shard buffer boundary is reached.

The implementation is also simpler: DispatchRuns precomputes the tensor byte layout, DirectMemoryWriter only walks that plan with a small cursor, and DirectShardWriter owns the per-shard DMA buffer rotation and PJRT transfer state and slop from the POC released alongside the zml/v2 is gone.

```
    std.Io.Reader
          |
          v
    DirectMemoryWriter.interface.buffer
          |
          | aliases primary shard DMA memory
          v
    +-----------------------------+
    | DirectShardWriter(primary)  |
    | current DMA pool buffer     |
    +-----------------------------+
          |
          | commitWindow()
          |   - advance byte_cursor
          |   - advance primary shard end
          |   - copy committed bytes to mirrors
          |   - flush at buffer/chunk boundaries
          v
    +-----------------------------+      +-----------------------------+
    | DirectShardWriter(primary)  |      | DirectShardWriter(mirror)   |
    | submitBuffered()            |      | writeAll() + submitBuffered |
    +-----------------------------+      +-----------------------------+
          |                                      |
          +------------------+-------------------+
                             v
                  PJRT transferData()
                             |
                             v
                  async event returns DMA buffer
```

### DIFF
--- a/zml/io.zig
+++ b/zml/io.zig
@@ -584,27 +584,30 @@ const DispatchSpans = struct {
         writer_index: usize,
         start: usize,
         len: usize,
+        order: usize,
     };
-
-    const PlacementSpans = std.MultiArrayList(PlacementSpan);
 
     spans: []DispatchSpan,
     mirror_writers: []usize,
 
     fn init(allocator: std.mem.Allocator, shape: Shape, sharding: Sharding) !DispatchSpans {
-        var placement_spans: PlacementSpans = .empty;
-        defer placement_spans.deinit(allocator);
-        var placement_span_order: std.ArrayList(usize) = .empty;
-        defer placement_span_order.deinit(allocator);
-
         const placement = try sharding.placement(shape);
+        const ordered_devices = sharding.devicesInCanonicalOrder();
+
+        var placement_span_count: usize = 0;
+        for (ordered_devices) |device| {
+            placement_span_count += placementSpanCount(shape, placement.slices(device.coords).constSlice());
+        }
+
+        var placement_spans: std.ArrayList(PlacementSpan) = try .initCapacity(allocator, placement_span_count);
+        defer placement_spans.deinit(allocator);
+
         const byte_strides = shape.computeByteStrides();
 
-        for (sharding.devicesInCanonicalOrder(), 0..) |device, writer_index| {
+        for (ordered_devices, 0..) |device, writer_index| {
             try appendShardPlacementSpans(
                 allocator,
                 &placement_spans,
-                &placement_span_order,
                 shape,
                 placement.slices(device.coords).constSlice(),
                 byte_strides.constSlice(),
@@ -612,12 +615,15 @@ const DispatchSpans = struct {
             );
         }
 
-        var spans: std.ArrayList(DispatchSpan) = .empty;
+        std.debug.assert(placement_spans.items.len == placement_span_count);
+
+        var spans: std.ArrayList(DispatchSpan) = try .initCapacity(allocator, placement_spans.items.len);
         errdefer spans.deinit(allocator);
-        var mirror_writers: std.ArrayList(usize) = .empty;
+
+        var mirror_writers: std.ArrayList(usize) = try .initCapacity(allocator, placement_spans.items.len);
         errdefer mirror_writers.deinit(allocator);
 
-        try deduplicateByRange(allocator, placement_spans, placement_span_order.items, shape.byteSize(), &spans, &mirror_writers);
+        try deduplicateByRange(allocator, placement_spans.items, shape.byteSize(), &spans, &mirror_writers);
 
         const spans_ = try spans.toOwnedSlice(allocator);
         errdefer allocator.free(spans_);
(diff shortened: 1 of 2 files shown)


zml/io.zig
```diff
@@ -434,75 +436,38 @@ const DirectShardWriter = struct {
...
         }
 
-        return consumed;
+        return chunk.len;
     }
 
     fn flush(w: *std.Io.Writer) std.Io.Writer.Error!void {
         const self: *DirectShardWriter = @alignCast(@fieldParentPtr("interface", w));
-        try self.flushBuffered();
+        try self.submitBuffered();
```
gwenzek: why the changed name ?

zml/io.zig
```diff
@@ -573,193 +535,260 @@ const DirectShardWriter = struct {
...
+                allocator,
+                &raw_runs,
+                &raw_order,
+                shape,
+                placement.slices(device.coords).constSlice(),
+                byte_strides.constSlice(),
+                writer_index,
+            );
         }
+        sortRawRuns(raw_runs, raw_order.items);
```
gwenzek: why do you need to sort "runs" ?

zml/io.zig
```diff
@@ -573,193 +535,260 @@ const DirectShardWriter = struct {
...
-                try self.segments.append(self.allocator, .{
-                    .len = first.len,
-                    .primary_writer = first.writer_index,
-                    .mirror_start = mirror_start,
-                    .mirror_len = j - i - 1,
-                });
-                cursor += first.len;
-                i = j;
-            }
+        try compactRawRuns(allocator, raw_runs, raw_order.items, shape.byteSize(), &items, &mirrors);
```
gwenzek: why do you need to "compact" ? could it be done during insertion ?

zml/io.zig
```diff
@@ -573,193 +535,260 @@ const DirectShardWriter = struct {
...
-            self.raw_segments.deinit(self.allocator);
-            self.segments.deinit(self.allocator);
-            self.mirrors.deinit(self.allocator);
-        }
+// Precomputed dispatch cursor data. Raw placement slices are emitted as global
+// tensor byte ranges, sorted, then compacted into final `[start, end)` runs.
+// Each run names the primary shard writer and a contiguous slice of `mirrors`
+// containing shard writers that must receive a copy of the same bytes.
+const DispatchRuns = struct {
+    const Run = struct {
```
gwenzek: I find those names confusing. What's a `Run` vs a `RawRun` ? what is a `Mirror` ?

gwenzek (commented):
I like where this is going , but I don't quite understand it

zml/io.zig
```diff
@@ -868,181 +886,118 @@ pub const DirectMemoryWriter = struct {
 
     fn drain(w: *std.Io.Writer, data: []const []const u8, splat: usize) std.Io.Writer.Error!usize {
         const self: *DirectMemoryWriter = @alignCast(@fieldParentPtr("interface", w));
-        var consumed: usize = 0;
+        if (splat != 1 or data.len != 1) return std.Io.Writer.Error.WriteFailed;
```
Corendos: I think you can just ignore instead of returning an error. If you only consume the first slice in `data` and return `n` accordingly, it will call you back with the rest later.

Corendos (approved):
Overall looks good but I didn't dive in the details

zml/io.zig
hugomano: Yes true!

zml/io.zig
hugomano: Yes reverted, `flushBuffered` is a better semantic.

zml/io.zig
hugomano: We sort because raw runs are emitted per device/shard, while the input reader streams bytes in global tensor order. Sorting turns placement-order runs into stream-order dispatch runs.

zml/io.zig
hugomano: After sorting, adjacent entries with the same `{start, len}` mean replicated / mirrored bytes:
```
raw: writer 0 owns [0, 1024)
raw: writer 1 owns [0, 1024)
```
Those should become one dispatch run:
```
primary_writer = 0
mirror_writers = [1]
range = [0, 1024)
```

Doing that during insertion is not cleanly with the current emission order. During insertion we are walking one device at a time, before global ordering is known. Mirrored ranges from another device may be inserted much later. After sorting, identical ranges are adjacent, so compaction is simple and deterministic.

zml/io.zig
hugomano: I agree with the review: `Run`, `RawRun`, and `Mirror` were vague.
I changed the naming to use span, since these objects represent global tensor byte intervals:
```
RawRun => PlacementSpan
Run => DispatchSpan
items => spans
mirrors => mirror_writers
```
Mental model:
`PlacementSpan`: byte span emitted from sharding placement traversal, in device order.
`DispatchSpan`: byte span consumed by `DirectMemoryWriter`, in global stream order.
`mirror_writers:` additional shard writers that receive the same byte span for replicated data.

I also reworked the comment above the struct to explain the full flow: placement traversal emits spans in device order, we sort them into reader order, then compact identical spans into one primary dispatch span plus mirror writers.

zml/io.zig
Finistere: Sounds like sort & compact would be better in a `deduplicateByRange`?

zml/io.zig
Finistere: Off the top of my head, maybe a `PlacementSpan` is a `Transfer` and `DispatchSpan` a `DeduplicatedTransfer` ?

zml/io.zig
```diff
@@ -573,193 +534,269 @@ const DirectShardWriter = struct {
...
 
+// Direct load writer state machine.
+//
+//     std.Io.Reader
+//          |
+//          v
+//     DirectMemoryWriter.interface.buffer
+//          | aliases active primary DirectShardWriter DMA buffer
+//          v
+//     commitWindow()
```
Finistere, line 763: nitpick: I would consider using `flushBuffered` like the ShardWriter for consistency, they're called for the same purpose, flushing the current interface.buffer.

zml/io.zig
hugomano: `deduplicateByRange` yes and done

Finistere (approved):

zml/io.zig
hugomano: Stay as it, we discussed about it.

zml/io.zig
hugomano: > Off the top of my head, maybe a `PlacementSpan` is a `Transfer` and `DispatchSpan` a `DeduplicatedTransfer` ?

Stay as it as discussed.

zml/io.zig
```diff
@@ -573,193 +534,264 @@ const DirectShardWriter = struct {
...
-        fn collectRawSegments(self: *StreamPlanner) !void {
-            const placement = try self.sharding.placement(self.shape);
-            for (self.sharding.devicesInCanonicalOrder(), 0..) |device, writer_index| {
-                try self.appendShardSegments(placement.slices(device.coords).constSlice(), writer_index);
-            }
-        }
+    fn init(allocator: std.mem.Allocator, shape: Shape, sharding: Sharding) !DispatchSpans {
+        var placement_spans: PlacementSpans = .empty;
+        defer placement_spans.deinit(allocator);
+        var placement_span_order: std.ArrayList(usize) = .empty;
```
gwenzek: consider adding an "order" field to PlacementSpan, rather than managing another arraylist like that.

gwenzek (commented):
Now that I understand it's pretty cool.

I would like to have smth similar for sharded read D2H , where we only read the necessary slices of data.

I left a comment to simplify a bit the code.

zml/io.zig
```diff
@@ -573,193 +534,264 @@ const DirectShardWriter = struct {
...
-                    if (lhs.start != rhs.start) return lhs.start < rhs.start;
-                    if (lhs.len != rhs.len) return lhs.len < rhs.len;
-                    return lhs.writer_index < rhs.writer_index;
-                }
-            }.lessThan);
+        const placement = try sharding.placement(shape);
+        const byte_strides = shape.computeByteStrides();
+
+        for (sharding.devicesInCanonicalOrder(), 0..) |device, writer_index| {
+            try appendShardPlacementSpans(
```
gwenzek: typically how many slices are we generating ? could you put some reasonable upper/lower bound so that we minimize reallocations in `appendShardPlacementSpans` ?

zml/io.zig
```diff
@@ -573,193 +534,264 @@ const DirectShardWriter = struct {
...
-
-                const mirror_start = self.mirrors.items.len;
-                var j = i + 1;
-                while (j < self.raw_segments.items.len and
-                    self.raw_segments.items[j].start == first.start and
-                    self.raw_segments.items[j].len == first.len) : (j += 1)
-                {
-                    try self.mirrors.append(self.allocator, self.raw_segments.items[j].writer_index);
-                }
+        var spans: std.ArrayList(DispatchSpan) = .empty;
```
gwenzek: I'm feeling inclined toward `.initCapacity(allocator, placement_spans.items.len)`

hugomano: Done. The final dispatch span count is bounded by the number of placement spans because deduplication only merges identical ranges, so `spans` is now initialized with `placement_spans.items.len`. I also preallocate `mirror_writers` with the same bound.

zml/io.zig
hugomano: Added an exact reserve pass before emission. For each shard, the number of emitted placement spans is the product of `slice.size` over axes before the contiguous suffix axis. This gives 1 span for common outer-axis sharding and row-count spans for innermost-axis sharding, matching the actual emitter behavior and avoiding append reallocations.

zml/io.zig
hugomano: Good point. I removed the extra `placement_span_order` list and folded insertion order into `PlacementSpan.order`. The builder now keeps a single temporary `ArrayList(PlacementSpan)`, sorts it directly by `{start, len, order}`, and uses `order` only as the stable tie-breaker.

zml/io.zig
gwenzek: cool I didn't realize that. If you trust yourself you can even remove the allocator from `appendShardPlacementSpans`

gwenzek (approved):
nice work !

zml/io.zig
hugomano: Yep, done. I removed the allocator/error plumbing from `appendShardPlacementSpans` and related helpers, and kept an assert that the emitted count matches the precomputed capacity.

PR #578, opened 2026-06-18
zml/io(xet): add xet support to VFS

oboulant:
This PR adds support for XET to the VFS. 

The XET machinery is in `zml/io/vfs/xet.zig`. 

`zml/io/vfs/lz4.zig` and `zml/io/vfs/bg4.zig` holds the decompression logic. 

In `zml/io/vfs/hf.zig`, we attach a `xet.FetchPool` to `HF` so that the worker group and the queue is re-used across calls to `HF.performRead()`. 

Modifications in `zml/io.zig` are unrelated to XET. Those are modifications done because of "bug" we surfaced during testing. They will be addressed in https://github.com/zml/zml/pull/576 

There are degrees of freedom: 
- `.dma_chunks`
- `.dma_chunk_size`
- `.parallelism`
- `XET_INTRA_TENSOR_WORKERS`: number of workers in the `std.Io.Group` 
- `.queue_capacity = XET_INTRA_TENSOR_WORKERS * 4`: the `*4` commands the capacity of the job queue

The memory footprint per worker:
- `body_size = max_xorb_body_size      // 64 * 1024 * 1024`: to hold downloaded data
- `scratch_size = 256 * 1024`: to hold a decompressed chunk. The maximum size of a chunk is supposed to be 128 KiB, but I have seen larger chunks in reality. 

The size are mentioned here: https://huggingface.co/docs/xet/xorb


zml/io/vfs/hf/bg4.zig
Corendos, line 1: This file should probably be moved to `zml/io/vfs/hf/bg4.zig`

zml/io/vfs/hf/lz4.zig
Corendos, line 1: This file should probably be moved to `zml/io/vfs/hf/lz4.zig`

zml/io/vfs/hf/xet.zig
Corendos, line 1: This file should probably be moved to `zml/io/vfs/hf/xet.zig`

zml/io/vfs/xet.zig
```diff
@@ -0,0 +1,645 @@
...
+/// Block the calling worker thread for `ms` milliseconds via libc.
+/// We don't have a `std.Io` handle in here and `std.Thread.sleep` is gone
+/// in Zig 0.16; this is the most direct portable replacement.
+fn sleepMs(ms: u64) void {
+    const ts: std.posix.timespec = .{
+        .sec = @intCast(ms / 1000),
+        .nsec = @intCast((ms % 1000) * std.time.ns_per_ms),
+    };
+    _ = std.c.nanosleep(&ts, null);
+}
```
Corendos: This is probably not required. If at some point you need to sleep and you don't have an `Io` available, you have two solutions:
* Bring an `Io` from above in the callstack
* If that's not possible, it's often because you are in a context where you are on a single thread and you can instantiate a `std.Io.Threaded` with `.init_single_threaded`.

In your case, at the end of the file, the context is a worker that runs in a group so I think you can find a way to inject the correct `io`.

zml/io/vfs/hf/xet.zig
```diff
@@ -0,0 +1,645 @@
...
+        body: []u8,
+        scratch: []u8,
+    };
+
+    pub fn init(
+        allocator: std.mem.Allocator,
+        io: std.Io,
+        client: *std.http.Client,
+        opts: Options,
+    ) !*FetchPool {
```
Corendos, line 493: Why does `init` return a `FetchPool` pointer ?

zml/io/vfs/hf/xet.zig
```diff
@@ -0,0 +1,645 @@
...
+        self.pending -= 1;
+        if (self.pending == 0) self.cond.signal(io);
+    }
+
+    fn wait(self: *Batch, io: std.Io) void {
+        self.mutex.lockUncancelable(io);
+        defer self.mutex.unlock(io);
+        while (self.pending != 0) self.cond.wait(io, &self.mutex) catch {};
+    }
+};
```
Corendos, line 571: I'm wondering if this could be replaced with already available primitives.

`fetchRange` could instantiate a `std.Io.Group` with tasks dispatching and each waiting for their respective job to finish on the `FetchPool`.

Could be more wasteful though, I was just being curious

zml/io/vfs/xet.zig
```diff
@@ -0,0 +1,645 @@
...
+    batch: *Batch,
+};
+
+fn fetchWorker(pool: *FetchPool, w: *FetchPool.Worker) void {
+    while (true) {
+        const job = pool.job_queue.getOne(pool.io) catch break;
+        if (job.batch.err != null) {
+            job.batch.finishOne(pool.io, null);
+            continue;
+        }
```
Corendos: Not a big deal but `batch.err` access could be subject to race since each job can write to it

zml/io/vfs/xet.zig
```diff
@@ -0,0 +1,645 @@
...
+
+        dst_off += wanted;
+        skip_head = 0;
+    }
+    return tasks;
+}
+
+pub const FetchPool = struct {
+    allocator: std.mem.Allocator,
+    io: std.Io,
```
Corendos: As a rule of thumb I think it's often better to not have struct own `std.Io` but function instead. Would it be possible to push the `io` to the `fetchRange` function ?

zml/io/vfs/xet.zig
```diff
@@ -0,0 +1,645 @@
...
+    dst: []u8,
+) !void {
+    if (dst.len == 0) return;
+
+    const cas = try cas_cache.get(pool.allocator, pool.io, pool.client, repo, hf_token);
+    defer pool.allocator.free(cas.url);
+    defer pool.allocator.free(cas.token);
+
+    var auth_buf: [4096]u8 = undefined;
+    const cas_auth = std.fmt.bufPrint(&auth_buf, "Bearer {s}", .{cas.token}) catch return error.TokenTooLong;
```
Corendos: This should probably use `bearerAuth`

Corendos (commented):
Overall I think that the implementation is good. I left some specific comment but more broadly:

* I feel like there could be a bit more factorization/cleaning to be done (the comment about `bearerAuth` is such an example).

zml/io/vfs/hf/xet.zig
oboulant: It returns a pointer because `FetchPool.init()` starts worker tasks immediatly, and those workers keep a `* FetchPool ` in `fn fetchWorker(pool: *FetchPool, w: *FetchPool.Worker) void {`.

Would you prefer it to be a two phase API: return `FetchPool ` by value unstarted, then start it later from a stable `*HF` / `*FetchPool`: `HF.init()` + `HF.start()` ? It would add a `HF.start()` though.

zml/io/vfs/hf/xet.zig
oboulant: I think it is possible. 

Current logic: `Batch` holds the countdown latch logic to wait for all the `FetchTask`s to complete on the long lived `FetchPool`. 

Also possible (I think): `fetchRange` would create its own `std.Io.Group` and  one task per job enqueued on `FetchPool` and then wait for that specific job to be signaled as done by a pool worker.  This could replace the batch-level wait, but we would still need per-job completion state/signaling.

What do you think ? 

My take home thinking: we need a logic to know all the specific `FetchTask`s needed to fulfill the `fetchRange()` call, submitted to the queue, are done. We can host this logic to a `Batch` struct (single `std.Io.Mutex`, `std.Io.Condition` and `err`) or handle it at the `fetchRange()` level. But we need some kind of logic to track this.

zml/io/vfs/hf/bg4.zig
oboulant: Done

zml/io/vfs/hf/lz4.zig
oboulant: Done

zml/io/vfs/hf/xet.zig
oboulant: Done

zml/io/vfs/xet.zig
oboulant: Thx !

I injected `io` down the  callstack. I deleted the `sleepMs()`.

zml/io/vfs/xet.zig
oboulant: Thx ! 

I added an atomic `Batch.failed` atomic flag that can be read without locking the `Batch.mutex`. 

WDYT ?

zml/io/vfs/xet.zig
oboulant: I removed `FetchPool.io` and passed it through the callstack.

zml/io/vfs/xet.zig
oboulant: Good catch ! Thx !

zml/io/vfs/hf/xet.zig
Corendos: Ah yes, it slipped through, I didn't see it

zml/io/vfs/hf/xet.zig
Corendos: Let's keep it that way

zml/io/vfs/xet.zig
Corendos: I think it's good

(There is still a potential weird wakeup ordering because we signal the condvar while still owning the lock, but that's no big deal I guess)

oboulant: Hmm, indeed the behavior is unclear. 

I have changed it so now we compute the condition under the lock, release the lock and finally signal on the condvar. 

What the doc says:
- https://github.com/ziglang/zig/blob/master/lib/std/Thread/Condition.zig is explicit but it is related to `std.Thread.Condition` 

> `signal()` can be called with or without the relevant Mutex being acquired 

- https://github.com/ziglang/zig/blob/master/lib/std/Io.zig does not say much

Bottom line, we should be fine with this modif that makes it explicit. 

@Corendos ⬆️

PR #583, opened 2026-06-19
zml/sharding : introduce shardableDim api

loupicaaa:
This PR introduces shardableDim helper in Sharding to handle the case where :
(dimension % number of sharding axes) != 0

Useful to stutter and shard correctly kv heads dim for example.

### DIFF
--- a/examples/llm/models/qwen3_5/model.zig
+++ b/examples/llm/models/qwen3_5/model.zig
@@ -536,8 +536,7 @@ pub const SelfAttn = struct {
         return .{ k, v };
     }
 
-    fn partitionProjectedKv(tensor: zml.Tensor, kv_head_sharding: zml.Sharding.DimSharding) zml.Tensor {
-        var kv = tensor;
+    fn partitionProjectedKv(kv: zml.Tensor, kv_head_sharding: zml.Sharding.DimSharding) zml.Tensor {
         return switch (kv_head_sharding) {
             .sharded => |heads| blk: {
                 if (heads.factor != 1) {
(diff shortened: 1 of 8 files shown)


examples/llm/models/qwen3_5/model.zig
```diff
@@ -873,11 +873,11 @@ pub const GatedDeltaNet = struct {
 
         const query_for_rule = b: {
             if (self.qk_head_repetition == 1) break :b query;
-            break :b query.stutter1d(@intCast(query.axis(.kh)), @intCast(self.qk_head_repetition));
+            break :b query.stutter1d(query.axis(.kh), self.qk_head_repetition);
```
gwenzek, line 874: why do you need to explicitly repeat the rows of key ?

hugomano: `query`/`key` are produced with `linear_num_key_heads` (`.kh`), while `value`, `g`, `beta`, and the recurrent state are indexed by `linear_num_value_heads` (`.vh`).

`recurrentGatedDeltaRule` renames `.kh` to `.vh` before calling `zml.nn.GatedDeltaNet.forward`, so q/k need the same head cardinality as value/state. The stutter is the GQA-style expansion: each q/k head is shared across `linear_num_value_heads / linear_num_key_heads` value heads.

So this is not repeating learned weight rows; it is expanding the activation head axis so the recurrence sees aligned `.h` dimensions. Without it, q/k have fewer heads than value/beta/g/state and the rule either shape-mismatches or pairs the wrong heads.

hugomano (approved):

PR #584, opened 2026-06-19
zml/build: bump bazel to 8.7.0

neudinger:


cerisier (approved):
Nothing major or interesting for us there apart from general improvements so LGTM :)

PR #586, opened 2026-06-19
zml/io: rework `zml.io.load`

Corendos:
The old `zml.io.load` was plagued with problems. This PR is an attempt at improving the state of things.

One major problem of `zml.io.load` was that it was not very composable. It was either using it to load all the model or doing everything by hand.

Another problem was that it didn't offer a way to apply transformations at load time to the tensors stored in a safetensors file. For example, it happens that for MoE models, instead of the experts weights to be tightly packed in a nice 3D tensor, the weight file contain one tensor for each expert. This is nice until your implementation required a 3D tensor as input. Using the old `zml.io.load` API this resulted in very convoluted way to do concatenation at load time. With this new API, we offer convenience so that it can be handled nicely and as fast as possible.

The main change is the introduction of the `Loader` struct. It's a struct that you instantiate once and reuse whenever you want to load tensors. It relies on `Binding`, which are a way to store which tensor are required and how they need to be combined (if it's a complex binding involving more than one source tensor).

One advantage of this `Loader` is that it's composable. Loading a struct using `Loader.auto(Struct)` or loading each individual field of `Struct` has essentially the same behavior. This solved the major problem of the old solution.

For simplicity I kept the old model loading code almost identical, as this new solution is a drop-in replacement.

# Update

After some discussions with @gwenzek I've changed my mind. Instead of having a declarative API, we'll provide helper to do transformation easily. The main changes that this PR now bring are:
* Introduce `Loader` struct
* `Loader` exposes `load` which is the composable replacement of `zml.io.load`
* `Loader` exposes `loadExecute` that allows you to execute an executable with the buffers loaded from the sources associated with a tensor. This allows for example to concatenate tensors.

We'll continue to work on it, this is just a preliminary work that allows us to do transformation at load time in LLMD.


examples/io/main.zig
```diff
@@ -200,21 +200,27 @@ pub fn main(init: std.process.Init) !void {
...
-                .total_bytes = &total_bytes,
             });
+            defer loader.deinit();
+
+            loader.auto(io, AllTensorsModel, &model, &buffers, &store, &.{sharded_sharding}, .{ .progress = &progress });
+            try loader.await(io);
+
+            const took = now.untilNow(io, .awake);
+            const total_bytes = loader.bytes_loaded.raw;
+            const bytes_per_sec: u64 = @intFromFloat(@as(f64, @floatFromInt(total_bytes)) / (@as(f64, @floatFromInt(took.nanoseconds)) / std.time.ns_per_s));
```
gwenzek, line 222: this would be clearer if all computation was done in u64. `totat_bytes * ns_per_s / took.ns` (also we have stdx.math.divFloat`)

examples/io/main.zig
```diff
@@ -200,21 +200,27 @@ pub fn main(init: std.process.Init) !void {
...
+            var loader: zml.io.Loader = try .init(allocator, platform, .{
                 .parallelism = 8,
                 .dma_chunks = 16,
                 .dma_chunk_size = 64 * zml.MiB,
-                .progress = &progress,
-                .total_bytes = &total_bytes,
             });
+            defer loader.deinit();
+
+            loader.auto(io, AllTensorsModel, &model, &buffers, &store, &.{sharded_sharding}, .{ .progress = &progress });
```
gwenzek: I will repeat it until I'm dead: "auto is a bad name". It's a computer, everything is automatic. 
`loader.loadAllAsync` would be better, or `loader.loadModelAsync`

gwenzek (commented):
TBH I don't like that you're pushing the concept of Binding inside the store.
I would have kept it loading side. 

I also don't like that it's declarative bindings, I would have preferred leveraging Loader apis directly from a `model.loadBuffer`.

You'll notice that it's more ceremonious to `unloadBuffer` than `loadBuffer`, even though the first one is decref in a for loop, and the other one has a lot of logic, choices, and interaction with the hardware.

I would allow custom `loadBuffer` and expose API to make concatenation / transformation easy in `loadBuffer`.

zml/io.zig
```diff
@@ -1071,127 +1190,229 @@ pub const LoadOpts = struct {
...
-    });
-    defer span.end();
+    dma_allocators: []const mem.DmaAllocator,
+    dma_chunk_size: usize,
+    pinned_buffer_pools: []mem.DynamicBufferPool,
+    group: stdx.Io.LimitedGroup,
+    bytes_loaded: std.atomic.Value(usize) = .init(0),
+
+    pub const Opts = struct {
+        pub const auto: Opts = .{
```
Finistere: nitpick: Wouldn't it be simpler to have default values? Or a `pub const default: Options` maybe?

zml/io.zig
```diff
@@ -1071,127 +1190,229 @@ pub const LoadOpts = struct {
...
+        };
+
+        meta.forEachVisit(model, *const Tensor, struct {
+            fn call(i: usize, tensor: *const Tensor, ctx_: *Ctx) void {
+                ctx_.self.group.async(ctx_.io, autoCallback, .{ ctx_.self, ctx_.io, tensor, ctx_.buffers[i], ctx_.store, ctx_.shardings, ctx_.opts });
+            }
+        }.call, .{&ctx});
+    }
+
+    fn autoCallback(self: *Loader, io: std.Io, tensor: *const Tensor, buffer: *Buffer, store: *const TensorStore, shardings: []const Sharding, opts: AutoOpts) void {
```
Finistere: `defaultCallback`?

examples/io/main.zig
Finistere: I would consider just using `load` for now, it's the only function to load a model anyway. Other cases, if any, might have better names

zml/io.zig
```diff
@@ -24,31 +24,33 @@ const log = std.log.scoped(.@"zml/io");
...
 
     pub fn deinit(self: *TensorStore) void {
         self.id_map.deinit(self.allocator);
+        self.arena.deinit();
     }
 
-    fn bindIdToKey(self: *TensorStore, key: []const u8, id: usize) !void {
-        const tensor_desc_ptr = self.registry.tensors.getPtr(key).?;
-
+    fn addBinding(self: *TensorStore, id: usize, binding: Binding) !void {
```
Finistere: nitpick: I would have named it `putUniqueBinding`

zml/io.zig
```diff
@@ -24,31 +24,33 @@ const log = std.log.scoped(.@"zml/io");
 
 pub const TensorStore = struct {
     registry: *safetensors.TensorRegistry,
-    id_map: std.AutoHashMapUnmanaged(usize, *safetensors.Tensor),
+    id_map: std.AutoHashMapUnmanaged(usize, Binding),
```
Finistere: nitpick: maybe `bindings`? or `id_to_binding`? `id_map` feels a bit too generic IMHO.

zml/io.zig
```diff
@@ -134,18 +142,28 @@ pub const TensorStore = struct {
         }
 
         pub fn maybeCreateTensor(self: View, subkey: []const u8, tagz: anytype, partitioning: anytype) ?Tensor {
-            var buffer: [256]u8 = undefined;
-            const key = std.fmt.bufPrint(&buffer, "{s}{s}", .{ self.prefix() orelse "", subkey }) catch unreachable;
+            return self.maybeCreateBinding(.{ .direct = subkey }, tagz, partitioning) catch unreachable;
```
Finistere: Not the only place I'm seeing this, but it's unclear why it's unreachable at all here. If it's truly unreachable, could you please add a `@panic("explain why this should be unreachable")` or be clearer with something like `@panic("error not handled")`?

Finistere (approved):

zml/io.zig
Corendos: What do you think of `putBindingNoClobber` to follow Zig's convention ?

zml/io.zig
Corendos: Yes, I kept the `.auto` we had before but I can rename to `.default`

zml/io.zig
Corendos: Addressed in https://github.com/zml/zml/pull/586/commits/2fd177c62fb9e94baafc0a9e275cca2e958bb939

zml/io.zig
Corendos: Addressed in https://github.com/zml/zml/pull/586/commits/b73b30a124593cc9d874c0ba9a0bcd4361e258c7

zml/io.zig
Corendos: Addressed in https://github.com/zml/zml/pull/586/commits/51c8ce6faac2adc05ad96f2fc12af0f711226b31

zml/io.zig
Corendos: Addressed in https://github.com/zml/zml/pull/586/commits/560a98ac8a584f34eac18828b96f54e31bb8647f

examples/io/main.zig
Corendos: Addressed in https://github.com/zml/zml/pull/586/commits/560a98ac8a584f34eac18828b96f54e31bb8647f

examples/io/main.zig
Corendos: Addressed in https://github.com/zml/zml/pull/586/commits/e2865df46e5307f2d502d151df829b2e1e366c80

zml/io.zig
Corendos: On second thought, `maybeCreateBinding` should not return an error so I just added `catch unreachable` everywhere in https://github.com/zml/zml/pull/586/commits/b73b30a124593cc9d874c0ba9a0bcd4361e258c7

Also, I agree that in general we should put explicit error message but in that case, it's not expected that it will fail and if that's the case the stack trace is enough to debug.

For example, in `putBindingNoClobber` I added a `@panic` because that can happen and it's useful for users. But would an allocation fail, I think it's not really useful to print an error message.

Finistere: If you consider that you don't want to handle allocation failure there, consider adding a `@panic("not handling allocation failure")` or something similar. That tells me what you're deliberately ignoring, and helps me if I ever need to touch that code in the future. `unreachable` is unhelpful when the `maybeCreateBinding` function has 70+ loc without any explicit error type.

Corendos: Ok, yeah makes sense. I'll improve that.

gwenzek: in general I use `@panic("message")` if I intend a user to read the message.
I try to reserve `unreachable` when you can prove that it's effectively unreachable. eg `if (x == 0) return; const y = std.math.divCeil(64, x) catch unreachable;` and prefer to leave a comment if non trivial.

gwenzek (approved):
thanks for the improvements, let's iterate on this API.

zml/io.zig
```diff
@@ -209,6 +295,263 @@ pub const TensorStore = struct {
...
+                        defer node.end();
+                        writer.setProgress(&node);
+                        defer writer.setProgress(null);
+                        var progress_writer: ProgressWriter = .init(writer.interface(), &node, .{ .scale = scale });
+
+                        var total: usize = 0;
+                        for (concatenate.tensors) |t| {
+                            var reader = t.reader(io, &.{}, .{}) catch unreachable;
+                            defer reader.deinit();
+                            total += reader.interface.streamRemaining(writer.interface()) catch unreachable;
```
elogir: shouldn't it be `&progress_writer.interface`?

Corendos: Good catch, addressed in https://github.com/zml/zml/pull/586/commits/71fef095fefdda606eb30f561a9ca740ab1365cc

examples/io/main.zig
Corendos: Actually I've reverted because large model where causing overflows

gwenzek (approved):

PR #587, opened 2026-06-19
attention: support `is_causal` option in Triton and Mosaic paged-attn kernels

tgautie:
We add suport for is_causal option in Triton and Mosaic paged-attention kernels, by adapting the mask-generation logic


Corendos (approved):

tgautie:
Note: this was tested to not add changes at the IR level in the default causal=true case, using the dsl-harness tool (from monorepo branch `raphael/dsl-harness-temp`).

PR #588, opened 2026-06-19
zml/attention: tune triton kernel for oneAPI performance

elogir:
Speeds up paged-attention decode on the Intel/oneAPI backend


elogir:
Note that I duplicated code on purpose in order not to mess with the working code for other platforms (until we fix the root cause that slows Triton / implement a sycl kernel)

Corendos (approved):

PR #591, opened 2026-06-22
third_party/iree: match all added tokens

elogir:
IREE's tokenizer only registered added tokens for pre-segmentation matching if they were flagged special or normalized. Tokens that are neither (e.g. <think> was dropped)

The fix is to match all added tokens


Finistere (approved):
nice

PR #594, opened 2026-06-23
zml/io/vfs(hf): introduce parallel read

oboulant:
This PR attaches to `HF` object a long lived `DownloadPool` field. 

`DownloadPool` holds a `std.Io.Queue` and starts stateless workers. Each job, when dequeued from the queue an assigned to a worker, is responsible to retrieve **at most** `default_download_chunk_size` bytes of data from Hugging Face. 

When downloaded, the data goes directly to its destination buffer (no intermediate storage). 

All requests on `performRead()` post jobs to the same queue.

`default_download_chunk_size`, `default_workers`, `default_queue_capacity` are degrees of freedom and affect performance. Chosen values for `dma_chunk_size` and `parallelism` from `zml.io.load()` also affect performances. 

---

# Tests

For:
- `const default_download_chunk_size = 10 * 1024 * 1024`, 
- `const default_workers = 16`, 
- `const default_queue_capacity = 64`
- `dma_chunk_size = 256 * zml.MiB`

#### Non sharded

```
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [4.135s]
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [4.286s]
info (llama): Loaded weights [14.96GiB, 18.457s, 829.81MiB/s]
```
#### Sharded

```
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [4.235s]
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [6.643s]
info (llama): Loaded weights [14.96GiB, 24.056s, 636.69MiB/s]
```

#### Non sharded

```
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [3.011s]
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [3.234s]
info (llama): Loaded weights [14.96GiB, 20.439s, 749.36MiB/s]
```
#### Sharded

```
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [3.053s]
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [4.874s]
info (llama): Loaded weights [14.96GiB, 27.842s, 550.11MiB/s]
```

#### Non sharded

```
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [3.544s]
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [3.638s]
info (llama): Loaded weights [14.96GiB, 20.64s, 742.07MiB/s]
```
#### Sharded

```
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [3.48s]
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [5.667s]
info (llama): Loaded weights [14.96GiB, 35.428s, 432.32MiB/s]
```
#### Non sharded

```
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [2.08s]
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [2.302s]
info (llama): Loaded weights [14.96GiB, 20.153s, 759.98MiB/s]
```
#### Sharded

```
info (zml/io): Loaded tensor lm_head.weight: 1050673152 bytes [2.712s]
info (zml/io): Loaded tensor model.embed_tokens.weight: 1050673152 bytes [4.674s]
info (llama): Loaded weights [14.96GiB, 31.47s, 486.69MiB/s]
```

In the same condition (please note there are some (large) network variability), total download time for `lm_head.weight` and `model.embed_tokens.weight` seems to indeed be faster when using the workers group attached to `HF`.

### DIFF
--- a/zml/io/vfs/hf.zig
+++ b/zml/io/vfs/hf.zig
@@ -51,8 +51,12 @@ const ParallelRead = struct {
         fn worker(pool: *Pool, io: std.Io) std.Io.Cancelable!void {
             while (true) {
                 const job = pool.job_queue.getOne(io) catch break;
-                job.result.* = job.perform(pool.client);
-                job.batch.finishOne(io);
+
+                if (!job.batch.failed()) {
+                    if (job.perform(pool.client)) |err| job.batch.fail(err);
+                }
+
+                job.batch.completeOne(io);
             }
         }
     };
@@ -62,8 +66,22 @@ const ParallelRead = struct {
         url: []const u8,
         authorization: std.http.Client.Request.Headers.Value,
         pending: std.atomic.Value(u32),
+        err: std.atomic.Value(u16) = .init(0),
 
-        fn finishOne(batch: *Batch, io: std.Io) void {
+        fn failed(batch: *Batch) bool {
+            return batch.err.load(.acquire) != 0;
+        }
+
+        fn fail(batch: *Batch, err: anyerror) void {
+            _ = batch.err.cmpxchgStrong(0, @intFromError(err), .release, .monotonic);
+        }
+
+        fn firstError(batch: *Batch) ?anyerror {
+            const err = batch.err.load(.acquire);
+            return if (err == 0) null else @errorFromInt(err);
+        }
+
+        fn completeOne(batch: *Batch, io: std.Io) void {
             const previous = batch.pending.fetchSub(1, .release);
             std.debug.assert(previous > 0);
             if (previous == 1) io.futexWake(u32, &batch.pending.raw, 1);
@@ -84,8 +102,6 @@ const ParallelRead = struct {
         chunk_offset: usize,
         chunk_len: usize,
         batch: *Batch,
-        err: ?anyerror = null,
-        result: *?anyerror,
 
         fn perform(job: Job, client: *std.http.Client) ?anyerror {
             var range_buf: [64]u8 = undefined;
@@ -917,17 +933,12 @@ pub const HF = struct {
                 .chunk_offset = chunk_offset,
                 .chunk_len = @min(self.read_pool.chunk_size, read_size - chunk_offset),
                 .batch = &batch,
-                .result = undefined,
             };
-            job.result = &job.err;
         }
 
         try self.read_pool.job_queue.putAll(self.base.inner, jobs);
         batch.waitUncancelable(self.base.inner);
-
-        for (jobs) |job| {
-            if (job.err) |err| return err;
-        }
+        if (batch.firstError()) |err| return err;
 
         return read_size;
     }
--- a/zml/io/vfs/hf.zig
+++ b/zml/io/vfs/hf.zig
@@ -8,7 +8,7 @@ const log = std.log.scoped(.@"zml/io/vfs/hf");
 
 const ParallelRead = struct {
     const Pool = struct {
-        pub const InitOptions = struct {
+        pub const InitOpts = struct {
             chunk_size: usize = 10 * 1024 * 1024,
             num_workers: usize = 16,
             queue_capacity: usize = 64,
@@ -20,7 +20,7 @@ const ParallelRead = struct {
         client: *std.http.Client,
         chunk_size: usize,
 
-        fn init(self: *Pool, allocator: std.mem.Allocator, io: std.Io, client: *std.http.Client, opts: InitOptions) !void {
+        fn init(self: *Pool, allocator: std.mem.Allocator, io: std.Io, client: *std.http.Client, opts: InitOpts) !void {
             stdx.debug.assert(opts.num_workers > 0, "Pool must have at least one worker", .{});
             stdx.debug.assert(opts.chunk_size > 0, "Pool must have a positive download chunk size", .{});
             stdx.debug.assert(opts.queue_capacity > 0, "Pool must have a positive queue capacity", .{});
@@ -234,7 +234,7 @@ const RepoKey = struct {
 
 pub const HF = struct {
     pub const InitOpts = struct {
-        read_pool: ParallelRead.Pool.InitOptions = .{},
+        read_pool: ParallelRead.Pool.InitOpts = .{},
     };
 
     pub const Repo = struct {
--- a/zml/io/vfs/hf.zig
+++ b/zml/io/vfs/hf.zig
@@ -6,6 +6,10 @@ const VFSBase = @import("base.zig").VFSBase;
 
 const log = std.log.scoped(.@"zml/io/vfs/hf");
 
+const default_download_chunk_size = 10 * 1024 * 1024;
+const default_workers = 16;
+const default_queue_capacity = 64;
+
 pub const API = struct {
     const TREE_URL_TEMPLATE = "https://huggingface.co/api/models/{[repo]s}/{[model]s}/tree/{[rev]s}/{[path]s}?expand=false&recursive=true&limit=1000";
     const LFS_FILE_URL_TEMPLATE = "https://huggingface.co/{[repo]s}/{[model]s}/resolve/{[rev]s}/{[path]s}";
@@ -50,6 +54,12 @@ const RepoKey = struct {
 };
 
 pub const HF = struct {
+    pub const InitOpts = struct {
+        download_chunk_size: usize = default_download_chunk_size,
+        workers: usize = default_workers,
+        queue_capacity: usize = default_queue_capacity,
+    };
+
     pub const Repo = struct {
         repo: []const u8,
         model: []const u8,
@@ -98,10 +108,91 @@ pub const HF = struct {
         }
     };
 
+    const DownloadPool = struct {
+        allocator: std.mem.Allocator,
+        client: *std.http.Client,
+        group: std.Io.Group = .init,
+        job_queue: std.Io.Queue(*Job),
+        queue_buf: []*Job,
+
+        fn init(allocator: std.mem.Allocator, io_: std.Io, client: *std.http.Client, opts: InitOpts) !*DownloadPool {
+            if (opts.workers == 0 or opts.download_chunk_size == 0 or opts.queue_capacity == 0) {
+                return error.InvalidDownloadPoolOptions;
+            }
+
+            const queue_buf = try allocator.alloc(*Job, opts.queue_capacity);
+            errdefer allocator.free(queue_buf);
+
+            const self = try allocator.create(DownloadPool);
+            errdefer allocator.destroy(self);
+
+            self.* = .{
+                .allocator = allocator,
+                .client = client,
+                .job_queue = .init(queue_buf),
+                .queue_buf = queue_buf,
+            };
+            errdefer self.group.cancel(io_);
+
+            for (0..opts.workers) |worker_id| {
+                try self.group.concurrent(io_, downloadWorker, .{ self, io_, worker_id });
+            }
+
+            return self;
+        }
+
+        fn deinit(self: *DownloadPool, io_: std.Io) void {
+            self.job_queue.close(io_);
+            self.group.await(io_) catch {};
+            self.allocator.free(self.queue_buf);
+            self.allocator.destroy(self);
+        }
+    };
+
+    const Batch = struct {
+        pending: usize,
+        failed: std.atomic.Value(bool) = .init(false),
+        err: ?anyerror = null,
+        mutex: std.Io.Mutex = .init,
+        cond: std.Io.Condition = .init,
+
+        fn finishOne(self: *Batch, io_: std.Io, err: ?anyerror) void {
+            self.mutex.lockUncancelable(io_);
+            if (err) |e| if (self.err == null) {
+                self.err = e;
+                self.failed.store(true, .release);
+            };
+            self.pending -= 1;
+            const done = self.pending == 0;
+            self.mutex.unlock(io_);
+            if (done) self.cond.signal(io_);
+        }
+
+        fn wait(self: *Batch, io_: std.Io) void {
+            self.mutex.lockUncancelable(io_);
+            defer self.mutex.unlock(io_);
+            while (self.pending != 0) {
+                self.cond.wait(io_, &self.mutex) catch {};
+            }
+        }
+    };
+
+    const Job = struct {
+        uri: []const u8,
+        authorization: std.http.Client.Request.Headers.Value,
+        data: []const []u8,
+        file_offset: u64,
+        dst_offset: usize,
+        len: usize,
+        batch: *Batch,
+    };
+
     allocator: std.mem.Allocator,
     mutex: std.Io.Mutex = .init,
     client: *std.http.Client,
     authorization: std.http.Client.Request.Headers.Value,
+    download_chunk_size: usize,
+    download_pool: *DownloadPool,
     handles: stdx.SegmentedList(Handle, 0) = .{},
     closed_handles: std.ArrayList(u32) = .empty,
     base: VFSBase,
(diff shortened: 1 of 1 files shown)


zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
+    const DownloadPool = struct {
+        allocator: std.mem.Allocator,
+        client: *std.http.Client,
+        group: std.Io.Group = .init,
+        job_queue: std.Io.Queue(*Job),
+        queue_buf: []*Job,
+
+        fn init(allocator: std.mem.Allocator, io_: std.Io, client: *std.http.Client, opts: InitOpts) !*DownloadPool {
+            if (opts.workers == 0 or opts.download_chunk_size == 0 or opts.queue_capacity == 0) {
+                return error.InvalidDownloadPoolOptions;
```
Finistere: I would consider just panicking here.

zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
+            for (0..opts.workers) |worker_id| {
+                try self.group.concurrent(io_, downloadWorker, .{ self, io_, worker_id });
+            }
+
+            return self;
+        }
+
+        fn deinit(self: *DownloadPool, io_: std.Io) void {
+            self.job_queue.close(io_);
+            self.group.await(io_) catch {};
```
Finistere: I would cancel here, if the `Pool` gets `.deinit()`, it means we don't care about the tasks anymore

zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
+
+        fn init(allocator: std.mem.Allocator, io_: std.Io, client: *std.http.Client, opts: InitOpts) !*DownloadPool {
+            if (opts.workers == 0 or opts.download_chunk_size == 0 or opts.queue_capacity == 0) {
+                return error.InvalidDownloadPoolOptions;
+            }
+
+            const queue_buf = try allocator.alloc(*Job, opts.queue_capacity);
+            errdefer allocator.free(queue_buf);
+
+            const self = try allocator.create(DownloadPool);
```
Finistere: Should be the caller that decides whether to alloacte on the heap (create) or on the stack

zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
+
+    const Batch = struct {
+        pending: usize,
+        failed: std.atomic.Value(bool) = .init(false),
+        err: ?anyerror = null,
+        mutex: std.Io.Mutex = .init,
+        cond: std.Io.Condition = .init,
+
+        fn finishOne(self: *Batch, io_: std.Io, err: ?anyerror) void {
+            self.mutex.lockUncancelable(io_);
```
Finistere: Are you sure about the uncancelable version? I wouldn't use them unless you really cannot tolerate cancellation.

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
     fn performRead(self: *HF, handle: *Handle, data: []const []u8, offset: u64) !usize {
         if (offset >= handle.size) return 0;
 
-        var range_buf: [64]u8 = undefined;
-        const range_header = blk: {
-            var total_bytes: u64 = 0;
-            for (data) |buf| {
-                total_bytes += @as(u64, buf.len);
+        const take = clampedReadLen(data, handle.size - offset);
```
Finistere: nitpick: Maybe `read_size`?

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
+        if (take == 0) return 0;
+
+        const job_count = jobCountForRead(take, self.download_chunk_size);
```
Finistere: nitpick: You could just do the divCeil here, the `jobCountFoRead` does the same `take == 0` check.

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
+                .file_offset = offset + dst_offset,
+                .dst_offset = dst_offset,
+                .len = @min(self.download_chunk_size, take - dst_offset),
+                .batch = &batch,
+            };
+        }
+
+        for (jobs, 0..) |*job, i| {
+            self.download_pool.job_queue.putOne(self.base.inner, job) catch |err| {
+                for (jobs[i..]) |_| batch.finishOne(self.base.inner, err);
```
Finistere: I'm not sure to understand why we need the finishOne / wait logic here. You seem to be just mutating the Batch state. `putOne` only fails if the queue is closed or the current future is cancelled, so I think a `try` is good enough.

I would consider adding a `errdefer batch.should_stop = true` instead to ensure we stop any jobs if this functions fails for some reason.

zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
+            self.job_queue.close(io_);
+            self.group.await(io_) catch {};
+            self.allocator.free(self.queue_buf);
+            self.allocator.destroy(self);
+        }
+    };
+
+    const Batch = struct {
+        pending: usize,
+        failed: std.atomic.Value(bool) = .init(false),
```
Finistere: I think you can skip the atomic here if the only purpose is to signal jobs in the `downloadWorker` to stop. I would rename it to something like `should_stop` instead.

Atomics are necessary when you want to ensure that some work done before/after is seen by other threads.

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
-            const end = offset + take - 1;
-            break :blk std.fmt.bufPrint(&range_buf, "bytes={d}-{d}", .{ offset, end }) catch unreachable;
-        };
+            const err = performJob(pool, job);
+            job.batch.finishOne(io_, err);
+        }
+    }
 
-        const repo: Repo = try .parse(handle.uri);
+    fn performJob(pool: *DownloadPool, job: *Job) ?anyerror {
```
Finistere: Maybe just `anyerror!void` and do a catch in `downloadWorker`?

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
+        downloadRange(pool.client, job) catch |err| return err;
+        return null;
+    }
+
+    fn downloadRange(client: *std.http.Client, job: *Job) !void {
+        const repo: Repo = try .parse(job.uri);
         var url_buffer: [8 * 1024]u8 = undefined;
         const url: []const u8 = try std.fmt.bufPrint(&url_buffer, API.LFS_FILE_URL_TEMPLATE, repo);
-
         const uri: std.Uri = try .parse(url);
```
Finistere, line 919: Could be done once and stored in the `Batch`

zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
+            }
+        }
+    };
+
+    const Job = struct {
+        uri: []const u8,
+        authorization: std.http.Client.Request.Headers.Value,
+        data: []const []u8,
+        file_offset: u64,
+        dst_offset: usize,
```
Finistere: nitpick: would consider using `chunk_offset` instead of `dst_offset` and `chunk_len` instead of just `len`

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
+        for (jobs, 0..) |*job, i| {
+            self.download_pool.job_queue.putOne(self.base.inner, job) catch |err| {
+                for (jobs[i..]) |_| batch.finishOne(self.base.inner, err);
+                batch.wait(self.base.inner);
+                if (batch.err) |e| return e;
+                return err;
+            };
+        }
+
+        batch.wait(self.base.inner);
```
Finistere: nitpick: You could have `wait` return an error if there was any and call it with `try`.

zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
+
+        fn finishOne(self: *Batch, io_: std.Io, err: ?anyerror) void {
+            self.mutex.lockUncancelable(io_);
+            if (err) |e| if (self.err == null) {
+                self.err = e;
+                self.failed.store(true, .release);
+            };
+            self.pending -= 1;
+            const done = self.pending == 0;
+            self.mutex.unlock(io_);
```
Finistere: Use `defer` instead

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
+
+    fn jobCountForRead(len: usize, chunk_size: usize) usize {
+        if (len == 0) return 0;
+        return std.math.divCeil(usize, len, chunk_size) catch unreachable;
+    }
+
+    fn downloadWorker(pool: *DownloadPool, io_: std.Io, worker_id: usize) void {
+        while (true) {
+            const job = pool.job_queue.getOne(io_) catch break;
+            log.warn("HF download worker {d:0>2} processing {s} bytes={d}-{d}, length={d}", .{
```
Finistere: Should be a `debug`

Finistere (approved):
nice

zml/io/vfs/hf.zig
Corendos: I think it's because the worker takes `self` and needs a stable pointer

zml/io/vfs/hf.zig
```diff
@@ -98,32 +108,129 @@ pub const HF = struct {
...
     base: VFSBase,
     trees: std.StringHashMapUnmanaged(std.ArrayList(TreeNode)) = .{},
     dir_read_states: std.AutoHashMapUnmanaged(*std.Io.Dir.Reader, ReadState) = .{},
 
     pub fn init(allocator: std.mem.Allocator, inner: std.Io, http_client: *std.http.Client, hf_token: ?[]const u8) !HF {
-        return .{
+        return initWithOpts(allocator, inner, http_client, hf_token, .{});
+    }
+
+    pub fn initWithOpts(allocator: std.mem.Allocator, inner: std.Io, http_client: *std.http.Client, hf_token: ?[]const u8, opts: InitOpts) !HF {
```
Corendos: I think we can just have `init` in that case

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
+    fn clampedReadLen(data: []const []u8, remaining: u64) usize {
+        var total: usize = 0;
+        for (data) |buf| total += buf.len;
+        return @intCast(@min(remaining, total));
+    }
+
+    fn jobCountForRead(len: usize, chunk_size: usize) usize {
+        if (len == 0) return 0;
+        return std.math.divCeil(usize, len, chunk_size) catch unreachable;
+    }
```
Corendos: As @Finistere this can probably be removed

zml/io/vfs/hf.zig
```diff
@@ -698,27 +806,93 @@ pub const HF = struct {
...
+                continue;
             }
-            const remaining = handle.size - offset;
-            const take = @min(remaining, total_bytes);
-            const end = offset + take - 1;
-            break :blk std.fmt.bufPrint(&range_buf, "bytes={d}-{d}", .{ offset, end }) catch unreachable;
-        };
+            const err = performJob(pool, job);
+            job.batch.finishOne(io_, err);
+        }
```
Corendos:
```suggestion
        while (pool.job_queue.getOne(io_)) |job| {
            log.debug("HF download worker {d:0>2} processing {s} bytes={d}-{d}, length={d}", .{
                worker_id,
                job.uri,
                job.file_offset,
                job.file_offset + job.len - 1,
                job.len,
            });
            if (job.batch.failed.load(.acquire)) {
                job.batch.finishOne(io_, null);
                continue;
}
            const err = performJob(pool, job);
            job.batch.finishOne(io_, err);
        } else |_| {}
```

https://ziglang.org/documentation/master/#while-with-Error-Unions

zml/io/vfs/hf.zig
Corendos: Would probably be more idiomatic yes

zml/io/vfs/hf.zig
Corendos: Agreed, and also maybe the worker could provide a `FixedBufferAllocator` to the job. That would remove the buffer on the stack (or the buffer could be in the stack of the worker)

zml/io/vfs/hf.zig
```diff
@@ -748,12 +922,36 @@ pub const HF = struct {
...
+    fn destSlice(data: []const []u8, dst_offset: usize, max_len: usize) ?[]u8 {
+        var cursor = dst_offset;
+        for (data) |buf| {
+            if (cursor >= buf.len) {
+                cursor -= buf.len;
+                continue;
+            }
+            return buf[cursor..][0..@min(max_len, buf.len - cursor)];
+        }
+        return null;
```
Corendos: My gut feeling is that there is a better way to do that 🤔 

Because here we go through `data` a lot (it's probably fine because it's in the cache though). Am I right to think that each call to `destSlice` will return a sub-slice of `[]u8` of `data` each time ?

Basically, depending on the total size we'll write, we'll use a sub-slice of `data` and fill all of them entirely except the first and last one.

Corendos: Maybe we should do it in the other way. Iterate on `data` instead of using the `remaining` and computing the index of `data` for each block

zml/io/vfs/hf.zig
```diff
@@ -6,6 +6,179 @@ const VFSBase = @import("base.zig").VFSBase;
...
+
+        fn init(allocator: std.mem.Allocator, io: std.Io, client: *std.http.Client, opts: InitOptions) !*Pool {
+            stdx.debug.assert(opts.num_workers > 0, "Pool must have at least one worker", .{});
+            stdx.debug.assert(opts.chunk_size > 0, "Pool must have a positive download chunk size", .{});
+            stdx.debug.assert(opts.queue_capacity > 0, "Pool must have a positive queue capacity", .{});
+
+            const queue_buf = try allocator.alloc(Job, opts.queue_capacity);
+            errdefer allocator.free(queue_buf);
+
+            const self = try allocator.create(Pool);
```
Finistere: I'm a bit confused by this pattern, IMHO the caller should decide where to put the Pool, wether on the heap or stack. Not `init()`

zml/io/vfs/hf.zig
```diff
@@ -698,79 +885,53 @@ pub const HF = struct {
...
 
-        return try reader.readSliceShort(data[0]);
-    }
-
-    const ContentRange = struct {
-        start: u64,
-        end: u64,
-        total: u64,
-    };
+        const queued = try self.read_pool.job_queue.put(self.base.inner, jobs, jobs.len);
```
Finistere: I think you want `putAll`. Currently, if you exceed the queue capacity of 64 you will fail instead of waiting.

zml/io/vfs/hf.zig
```diff
@@ -6,6 +6,179 @@ const VFSBase = @import("base.zig").VFSBase;
...
+        authorization: std.http.Client.Request.Headers.Value,
+        pending: std.atomic.Value(u32),
+
+        fn finishOne(batch: *Batch, io: std.Io) void {
+            const previous = batch.pending.fetchSub(1, .release);
+            std.debug.assert(previous > 0);
+            if (previous == 1) io.futexWake(u32, &batch.pending.raw, 1);
+        }
+
+        fn wait(batch: *Batch, io: std.Io) void {
```
Finistere: nitpick: I would rename it `waitUncancelable`

Finistere (approved):

PR #597, opened 2026-06-24
zml/safetensors: limit reads to 1GB

elogir:
On macOS, pread returns INT_MAX when the requested size is larger than 2GB

### DIFF
--- a/zml/safetensors.zig
+++ b/zml/safetensors.zig
@@ -128,13 +128,17 @@ pub const TensorReader = struct {
         self.file.close(self.io);
     }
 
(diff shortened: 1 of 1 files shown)


zml/safetensors.zig
```diff
@@ -128,13 +128,15 @@ pub const TensorReader = struct {
         self.file.close(self.io);
     }
 
+    const max_single_read_bytes: u64 = 1 << 30;
```
hugomano, line 133: leverage zig std please instead of hardcoding the value

hugomano (changes requested):

hugomano (approved):

PR #598, opened 2026-06-24
platforms/metal: sandbox Metal toolchain

elogir:
Fetch Apple's encrypted Metal toolchain and only exposes the minimal compile tools + runtime needed

### DIFF
--- a/platforms/metal/libpjrt_metal.BUILD.bazel
+++ b/platforms/metal/libpjrt_metal.BUILD.bazel
@@ -5,6 +5,7 @@ copy_file(
     name = "libpjrt_metal.dylib",
     src = "libpjrt_c_api_gpu_plugin.dylib",
     out = "lib/libpjrt_c_api_gpu_plugin.dylib",
+    allow_symlink = True,
 )
 
 copy_to_directory(
@@ -13,8 +14,6 @@ copy_to_directory(
         ":libpjrt_metal.dylib",
     ],
     add_directory_to_runfiles = False,
-    allow_overwrites = True,
-    hardlink = "auto",
     include_external_repositories = ["**"],
 )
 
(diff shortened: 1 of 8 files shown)


platforms/metal/metal_toolchain.BUILD.bazel
```diff
@@ -0,0 +1,382 @@
+package(default_visibility = ["//visibility:public"])
+
+# Static BUILD file for the 17B54 MetalToolchain asset.
+
+filegroup(
+    name = "toolchain_runtime",
+    srcs = glob(["Metal.xctoolchain/**"]),
+)
+
+exports_files([
```
cerisier, line 10: nit we don't use many of those files.

platforms/metal/libpjrt_metal.BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+)
+
+copy_to_directory(
+    name = "sandbox",
+    srcs = [
+        ":libpjrt_metal.dylib",
+    ],
+    add_directory_to_runfiles = False,
+    allow_overwrites = True,
+    hardlink = "auto",
```
cerisier: usually no need

platforms/metal/libpjrt_metal.BUILD.bazel
```diff
@@ -0,0 +1,25 @@
+load("@bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+
+copy_file(
+    name = "libpjrt_metal.dylib",
+    src = "libpjrt_c_api_gpu_plugin.dylib",
+    out = "lib/libpjrt_c_api_gpu_plugin.dylib",
+)
```
cerisier, line 9: use allow_symlink=True,

as a rule of thumb, all copy_file should be allow_symlink=True

cerisier (approved):

cerisier:
Merging as is, a follow up PR will cleanup the dylib with install name tool.

PR #618, opened 2026-07-01
zml: introduce zml.Compiler

gwenzek:
Move the type out of `zml.compile` signature so that we can call `io.async` more easily.

Sample code:

```
var layer_compile_future = io.async(zml.Compiler(TransformerLayer.forward).compile, .{
    allocator,
    io,
    platform,
    layer_compile_opts,
    .{
        model.model.layers[0],
        hidden,
        device_state.kv_cache.layers[0],
    },
});
```


Finistere (approved):
nice

Corendos (approved):

PR #631, opened 2026-07-05
stdx/json: fix potential invalid free

Corendos:
When we parsed from a std.json.Value we weren't correctly duping the strings, potentially leading to invalid free.

### DIFF
--- a/stdx/json.zig
+++ b/stdx/json.zig
@@ -45,8 +45,14 @@ pub fn Union(comptime T: type) type {
                 switch (field.type) {
                     bool => if (source == .bool) return .{ .value = @unionInit(T, field.name, source.bool) },
                     []const u8 => switch (source) {
-                        .string => |v| return .{ .value = @unionInit(T, field.name, v) },
-                        .number_string => |v| return .{ .value = @unionInit(T, field.name, v) },
+                        .string => |v| {
+                            const dupe = try allocator.dupe(u8, v);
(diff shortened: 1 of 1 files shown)


stdx/json.zig
```diff
@@ -45,8 +45,14 @@ pub fn Union(comptime T: type) type {
                 switch (field.type) {
                     bool => if (source == .bool) return .{ .value = @unionInit(T, field.name, source.bool) },
                     []const u8 => switch (source) {
-                        .string => |v| return .{ .value = @unionInit(T, field.name, v) },
-                        .number_string => |v| return .{ .value = @unionInit(T, field.name, v) },
+                        .string => |v| {
+                            const dupe = try allocator.dupe(u8, v);
```
gwenzek, line 49: we should respect the json parse options, it has a "should_allocate" field

Corendos: Yes, I've been down this rabbit hole. From my understanding this is not required in that case, it's only when you parse from a token source (because in that case pointers can be invalidated). If you parse from a `std.json.Value` it's different (but maybe I misunderstood)

gwenzek: the code source of `https://ziglang.org/documentation/master/std/#std.json.static.innerParseFromValue` also always copy strings, so let's do it the same way.

gwenzek (approved):

PR #635, opened 2026-07-07
zml/vfs: treat scheme:// paths as absolute in the router

XciD:
## The bug

In the VFS router, a scheme-qualified path like `hf://owner/model/file` is meant to be absolute: it should resolve from the scheme's root. But `lookupDir` only treated it as absolute when the base directory was `cwd`.

When such a path is opened *relative to an already-open directory* of the same backend, the router kept the directory's path and glued the new path after it, duplicating it. So `hf://owner/model` + `hf://owner/model/file` became `owner/model/owner/model/file`, and the lookup failed with `FileNotFound`.

## Why it matters

That is exactly the pattern `safetensors.fromPath` uses: it opens the model directory (`hf://owner/model`), then opens the full file path (`hf://owner/model/file.safetensors`) under it. The doubled path meant **no `hf://` model could be loaded** through `zml.io`.

## The fix

In `lookupDir`, any `scheme://` sub-path resolves from the scheme's root regardless of the base directory, the same way an absolute POSIX path already does. About four lines.

### DIFF
--- a/zml/io/vfs.zig
+++ b/zml/io/vfs.zig
@@ -137,6 +137,20 @@ pub const VFS = struct {
     }
 
     fn lookupDir(self: *VFS, dir: std.Io.Dir, sub_path: ?[]const u8) !struct { ?usize, std.Io.Dir, std.Io } {
+        // A scheme-qualified path (e.g. "hf://owner/model/file") is absolute: it
+        // is resolved from the scheme's root regardless of `dir`. Without this,
+        // opening such a path relative to an already-open dir of the same
+        // backend double-prefixes the dir's path.
+        if (sub_path) |sp| {
+            if (std.mem.indexOf(u8, sp, "://") != null) {
+                const uri = std.Uri.parse(sp) catch return error.VFSNotRegistered;
+                const backend_idx: usize = for (self.backends.entries.items(.key), 0..) |s, idx| {
+                    if (std.mem.eql(u8, uri.scheme, s)) break idx;
+                } else return error.VFSNotRegistered;
+                return .{ backend_idx, std.Io.Dir.cwd(), self.getBackend(backend_idx) };
+            }
+        }
+
         if (std.meta.eql(dir, std.Io.Dir.cwd())) {
             if (sub_path == null) return .{ null, dir, self.base.inner };
             if (std.fs.path.isAbsolutePosix(sub_path.?)) return .{ null, dir, self.base.inner };


zml/io/vfs.zig
```diff
@@ -137,6 +137,20 @@ pub const VFS = struct {
...
+        // backend double-prefixes the dir's path.
+        if (sub_path) |sp| {
+            if (std.mem.indexOf(u8, sp, "://") != null) {
+                const uri = std.Uri.parse(sp) catch return error.VFSNotRegistered;
+                const backend_idx: usize = for (self.backends.entries.items(.key), 0..) |s, idx| {
+                    if (std.mem.eql(u8, uri.scheme, s)) break idx;
+                } else return error.VFSNotRegistered;
+                return .{ backend_idx, std.Io.Dir.cwd(), self.getBackend(backend_idx) };
+            }
+        }
```
hugomano: Thanks, well seen. I have a small suggestion/nit: rooted paths should be resolved before considering the current dir. Both POSIX absolute paths and scheme-qualified VFS paths have “start from root” semantics, so grouping them here makes the lookup invariant clearer.

```suggestion
        if (sub_path) |sp| rooted_path: {
            if (std.fs.path.isAbsolutePosix(sp)) return .{ null, .cwd(), self.base.inner };

            const uri = std.Uri.parse(sp) catch break :rooted_path;
            if (!std.mem.startsWith(u8, sp[uri.scheme.len..], "://")) break :rooted_path;

            const backend_idx = for (self.backends.entries.items(.key), 0..) |scheme, idx| {
                if (std.mem.eql(u8, uri.scheme, scheme)) break idx;
            } else return error.VFSNotRegistered;
            return .{ backend_idx, .cwd(), self.getBackend(backend_idx) };
        }
```

zml/io/vfs.zig
```diff
@@ -137,6 +137,20 @@ pub const VFS = struct {
...
+
         if (std.meta.eql(dir, std.Io.Dir.cwd())) {
             if (sub_path == null) return .{ null, dir, self.base.inner };
             if (std.fs.path.isAbsolutePosix(sub_path.?)) return .{ null, dir, self.base.inner };
```
hugomano: So, we can remove this line.

hugomano (changes requested):
Thanks we appreciate the PR, just some nits then this we will merge!

zml/io/vfs.zig
```diff
@@ -137,20 +137,26 @@ pub const VFS = struct {
...
+            if (std.fs.path.isAbsolutePosix(sp)) return .{ null, .cwd(), self.base.inner };
+
+            const uri = std.Uri.parse(sp) catch break :rooted_path;
+            if (!std.mem.startsWith(u8, sp[uri.scheme.len..], "://")) break :rooted_path;
+
+            const backend_idx = for (self.backends.entries.items(.key), 0..) |scheme, idx| {
+                if (std.mem.eql(u8, uri.scheme, scheme)) break idx;
+            } else return error.VFSNotRegistered;
+            return .{ backend_idx, .cwd(), self.getBackend(backend_idx) };
+        }
```
hugomano, line 152: Sorry for the back and forth here. I discovered a bug with the suggestion I did through a manual test that exposed an important behavior difference.

I think we should keep the existing cwd branch behavior below and make the new early block handle only scheme-qualified VFS paths. So like what you did first..

In particular, I’d avoid moving this into the early `rooted_path` block:
`if (std.fs.path.isAbsolutePosix(sp)) return .{ null, .cwd(), self.base.inner };`

For an already-open backend dir, paths like `/adapters/foo` may still be intended for that backend. HF and co currently interpret leading `/` this way via `resolvePath`, so routing POSIX-absolute-looking subpaths to the base filesystem changes existing behavior.

I’d suggest reverting the removed cwd branch starting from:
`if (std.meta.eql(dir, std.Io.Dir.cwd())) {`
and keeping the new early logic scoped to scheme://... paths only. That preserves the fix for full VFS paths like `hf://owner/model/file`, without changing backend-relative `/foo` behavior.

hugomano (changes requested):
Sorry for back and forth we should only insert the block you did first.

zml/io/vfs.zig
XciD: no prob, I've revert it back

hugomano (approved):
Thanks @XciD

PR #641, opened 2026-07-09
zml/attention: implement batching in cu_fa2 backend

gwenzek:


zml/attention.zig
gwenzek, line 1: moved from attention/attention it's the last commit, so it's easy to review without this commit, but the change is just the new test cases.

zml/ops.zig
```diff
@@ -307,42 +307,51 @@ pub const ReduceWindowOpts = struct {
     padding: []const [2]i64,
 };
 
-pub fn reduceWindow(inputs: anytype, inits: anytype, opts: ReduceWindowOpts, comptime func: anytype, context: anytype) stdx.meta.FnReturn(func) {
```
gwenzek, line 310: originally I wanted to add a new cumsum to match the semantics needed by cu_fa2 (cumSumWithLeadingZero) )but in the end I didn't used it. 

So I drop the new cumsum, but kept this changes to reduceWindow that remove the needs for anytype. I'm not handling context, but if one day we need to, a `*anyopaque` is probably better than an `anytype`

zml/platform.zig
```diff
@@ -572,8 +572,15 @@ pub const Platform = struct {
...
     /// Memory is owned by the platform, making it safe to copy around.
-    pub fn registerSharding(platform: *Platform, name: []const u8, logical: Sharding.LogicalMesh) !Sharding {
-        return platform.registerShardingWithStrategy(name, logical, .suggest(logical, &platform.physical_mesh));
+    pub fn registerSharding(platform: *Platform, name: []const u8, logical: Sharding.LogicalMesh) error{OutOfMemory}!Sharding {
+        return platform.registerShardingWithStrategy(
+            name,
+            logical,
+            .suggest(logical, &platform.physical_mesh),
+        ) catch |err| switch (err) {
+            error.InvalidPhysicalMesh, error.InvalidStrategy, error.InvalidPhysicalAxis => unreachable,
```
gwenzek: those errors aren't possible to handle by the caller: we created the mesh and the strategy, so they must correspond.

zml/platform.zig
```diff
@@ -826,7 +833,7 @@ fn printCallbackInner(call_frame: *pjrt.ffi.CallFrame) !?*pjrt.ffi.Error {
     const slice: zml.Slice = .init(shape, host_visible_data[0..shape.byteSize()]);
     const name = call_frame.attrs.getByName(.string, "name").?.slice();
 
-    log.info("{s} {f} [device={d}]: {d}", .{ name, slice.shape, device_ordinal, slice });
+    std.debug.print("{s} {f} [device={d}]: {d}\n", .{ name, slice.shape, device_ordinal, slice });
```
gwenzek, line 838: I kinda like having tensor.print always work, even in tests. Let me know if you think otherwise.

zml/shape.zig
```diff
@@ -573,12 +573,13 @@ pub const Shape = struct {
...
-        new_shape.inferMissingAxis(self.count());
-        stdx.debug.assert(self.count() == new_shape.count(), "Can't reshape {any} to {any}", .{ self.dims(), new_shape.dims() });
+        new_shape.inferMissingAxis(self.count()) catch |err| {
+            std.debug.panic("Can't reshape {any} to {any}: {t}", .{ self.dims(), new_shape.dims(), err });
+        };
         return new_shape;
     }
 
-    fn inferMissingAxis(self: *Shape, n_: usize) void {
+    fn inferMissingAxis(self: *Shape, n_: usize) error{ UnexpectedRemainder, TooBig }!void {
```
gwenzek, line 582: return an error so that `reshape` and `splitAxis` can print appropriate error message.

examples/llm/models/lfm2/model.zig
```diff
@@ -213,8 +213,8 @@ pub const Model = struct {
         actual_seq_len: zml.Tensor,
         rng: zml.Tensor.Rng,
         cache_: Cache,
-        attention_metadata: zml.attention.attention.Metadata,
```
gwenzek, line 216: > you had my attention.attention, but now you have my attention.

zml/platform.zig
Finistere: Could you write it down in a `@panic` instead of just `unreachable` please?

zml/platform.zig
Finistere: 👍 I would also have expected tensor.print to... print and not log.

Finistere (approved):
nice

zml/platform.zig
gwenzek: I put a `std.debug.panic` with an error message inviting to open a bug.

PR #642, opened 2026-07-09
zml/tensor: fix IRFFT output shape

ousamabenyounes:
## Summary

Fix IRFFT shape inference so it uses the requested FFT output length and validates that the complex input length matches the corresponding RFFT output length.

Fixes #411.

## Why

RFFT changes the innermost real dimension `n` into `n / 2 + 1` complex bins. IRFFT receives those complex bins plus the requested real FFT length, so its output shape should restore that requested length instead of deriving a smaller value from the complex input dimension.

## Test verification (RED -> GREEN)

Command used for all runs:

```bash
npx -y @bazel/bazelisk@1.25.0 test //zml:test --test_output=errors
```

RED, with the new regression test before the fix:

```text
102/139 tensor.Tensor.test.Tensor.fft IRFFT restores RFFT last dimension...thread 1701155 panic: exact division produced remainder
INFO: Build completed, 1 test FAILED, 5355 total actions
//zml:test                                                               FAILED in 24.9s
Executed 1 out of 1 test: 1 fails locally.
```

GREEN, with the fix applied:

```text
INFO: Build completed successfully, 6 total actions
//zml:test                                                               PASSED in 8.1s
Executed 1 out of 1 test: 1 test passes.
```

Revert RED, with the production fix reverted while keeping the test:

```text
102/139 tensor.Tensor.test.Tensor.fft IRFFT restores RFFT last dimension...thread 1711292 panic: fft(.{ .kind = .IRFFT, .length = { 10 } }) expects tensor last dimensions to match given lengths, got 1 and 1
INFO: Build completed, 1 test FAILED, 6 total actions
//zml:test                                                               FAILED in 22.9s
Executed 1 out of 1 test: 1 fails locally.
```

## Additional checks

```text
zig fmt --check zml/tensor.zig
fmt_status=0
```

### DIFF
--- a/zml/tensor.zig
+++ b/zml/tensor.zig
@@ -559,24 +559,59 @@ pub const Tensor = struct {
                     .f32 => .c64,
                     else => .c128,
                 };
-                const shape_ = self._shape.setDim(-1, @divExact(self.dim(-1), 2) + 1);
+                const shape_ = self._shape.setDim(-1, rfftOutputLen(self.dim(-1)));
                 break :blk shape_.withDtype(dt);
             },
             .IRFFT => blk: {
                 stdx.debug.assert(self.dtype().isComplex(), "fft({any}) expects tensor type to be complex, got {}", .{ opts, self.dtype() });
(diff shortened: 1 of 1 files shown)


gwenzek (approved):
Thanks

PR #645, opened 2026-07-09
bin/zml-smi: fix memory & compute over-reporting for oneapi

catwell:
before:
before
after:
after

@neudinger

### DIFF
--- a/bin/zml-smi/platforms/oneapi/monitor.zig
+++ b/bin/zml-smi/platforms/oneapi/monitor.zig
@@ -243,7 +243,7 @@ pub fn collectProcessUsage(allocator: std.mem.Allocator, io: std.Io, devices: []
     var result: ProcessUsage = .{};
     errdefer result.deinit(allocator);
 
-    var seen_device_clients: std.AutoHashMapUnmanaged(u128, void) = .{};
+    var seen_device_clients: std.AutoHashMapUnmanaged(struct { u16, u64 }, void) = .{};
     defer seen_device_clients.deinit(allocator);
 
     var proc_dir = try std.Io.Dir.openDirAbsolute(io, "/proc", .{ .iterate = true });
@@ -270,7 +270,7 @@ pub fn collectProcessUsage(allocator: std.mem.Allocator, io: std.Io, devices: []
             // Skip fd if DRM client has been seen for this device.
             // Summing after dup() / fork() over-reports memory and compute.
             if (parsed.drm_client_id) |cid| {
-                if (try markDrmClientSeen(allocator, &seen_device_clients, dev_idx, cid)) continue;
+                if ((try seen_device_clients.getOrPut(allocator, .{ dev_idx, cid })).found_existing) continue;
             }
 
             const key = processKey(pid, dev_idx);
@@ -671,19 +671,6 @@ fn processKey(pid: u32, device_idx: u16) u64 {
     return (@as(u64, device_idx) << 32) | pid;
 }
 
-inline fn drmClientKey(device_idx: u16, client_id: u64) u128 {
-    return (@as(u128, device_idx) << 64) | @as(u128, client_id);
-}
-
-fn markDrmClientSeen(
-    allocator: std.mem.Allocator,
-    seen_device_clients: *std.AutoHashMapUnmanaged(u128, void),
-    device_idx: u16,
-    client_id: u64,
-) !bool {
-    return (try seen_device_clients.getOrPut(allocator, drmClientKey(device_idx, client_id))).found_existing;
-}
-
 pub const BdfParts = struct {
     domain: u32,
     bus: u32,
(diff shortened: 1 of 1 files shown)


neudinger:
Hello @catwell thanks for your contribution.

Thank you for using zml with zml-smi and testing on intel platform.


You only commented with a screenshot which shows only the outcome once the patch is applied, but you did not explain the problem and therefore what the solution was. \
After reviewing the PR change, I agree that the fix is correct and addresses the over-reporting issue.

After reading the change the over-reporting issue is :

If a process duplicated a GPU FD with dup() or inherited it through fork(), the same oneAPI DRM client could appear more than once.

In your case you use a intel cpu who also has an integrated GPU, and you have a process that uses the integrated GPU and the discrete GPU. \
The process duplicated the GPU FD and therefore the same oneAPI DRM client appears more than once in the /proc/<pid>/fdinfo/* files. \
The zml-smi tool sums the GPU memory and engine usage from the DRM fdinfo entries, and therefore it over-reports the GPU memory and engine usage.

```log
/proc/4242/fdinfo/10:
drm-pdev:         0000:03:00.0
drm-client-id:    42
drm-total-memory: 2048 MiB
drm-engine-compute: 300000000 ns

/proc/4242/fdinfo/11:
drm-pdev:         0000:03:00.0
drm-client-id:    42
drm-total-memory: 2048 MiB
drm-engine-compute: 300000000 ns
```

These are two FDs for the same DRM client, not two independent GPU user (client/handle).
Without it, zml-smi would sum both and report:

```log
GPU 0, drm-client-id 42 -> counted
GPU 0, drm-client-id 42 again -> counted
```

It should be counted only once:

```log
GPU 0, drm-client-id 42 -> counted
GPU 0, drm-client-id 42 again -> skipped
```

```log
PID 1000 fd 10, drm-client-id 42, memory 2 GiB
PID 1000 fd 11, drm-client-id 42, memory 2 GiB
```
Should report 2 GiB, not 4 GiB, because both FDs refer to the same underlying GPU client.

The fdinfo dedup approach looks right to me: duplicated/inherited DRM fds can expose the same DRM client multiple times, so summing per-fd over-reports memory and engine time.

Due to the possibility to have multiple GPUs or CPU with integrated graphics (NUMA node), I also propose adding one small robustness tweak locally should be `(device_idx, drm_client_id)` instead of just `drm_client_id`.
This avoids accidentally dropping usage from a second GPU that happens to report the same client ID.

I also propose adding a focused unit test covering:
- same device + same client ID => duplicate, skipped
- different device + same client ID => distinct, counted
- same device + different client ID => distinct, counted


Verified with:

`bazel test --cache_test_results=no //bin/zml-smi/platforms/oneapi:test`

If you are okay with the changes I pushed everything else looks good to me and can be merged.


Next time please provide a description of the problem and the solution in your PR comment, not just a screenshot of the result. \
It helps reviewers understand the context and reasoning behind the change. \
Of course screenshots are welcome, but they should be accompanied by a description of the problem and the solution.


Thanks again for your contribution, and for using **zml**.

catwell:
@neudinger Thanks! And sorry for the lack of description. I thought the comment + small enough diff was enough but that was not the case - TBH I rushed to push this because it was running late.

You are right regarding multiple devices! I think you could have just done this if you wanted minimum changes but what you did works too.

```diff
diff --git a/bin/zml-smi/platforms/oneapi/monitor.zig b/bin/zml-smi/platforms/oneapi/monitor.zig
index 7d826f31..913c09a3 100644
--- a/bin/zml-smi/platforms/oneapi/monitor.zig
+++ b/bin/zml-smi/platforms/oneapi/monitor.zig
@@ -243,7 +243,7 @@ pub fn collectProcessUsage(allocator: std.mem.Allocator, io: std.Io, devices: []
     var result: ProcessUsage = .{};
     errdefer result.deinit(allocator);

-    var seen_clients: std.AutoHashMapUnmanaged(u64, void) = .{};
+    var seen_clients: std.AutoHashMapUnmanaged(struct { u16, u64 }, void) = .{};
     defer seen_clients.deinit(allocator);

     var proc_dir = try std.Io.Dir.openDirAbsolute(io, "/proc", .{ .iterate = true });
@@ -267,9 +267,10 @@ pub fn collectProcessUsage(allocator: std.mem.Allocator, io: std.Io, devices: []
             const bus_device_identifier = parsed.bus_device_identifier orelse continue;
             const dev_idx = matchDevice(devices, bus_device_identifier) orelse continue;

-            // Skip fd if DRM client has been seen. Summing after dup() / fork() over-reports memory and compute.
+            // Skip fd if DRM client ID has been seen on same device.
+            // Summing after dup() / fork() over-reports memory and compute.
             if (parsed.drm_client_id) |cid| {
-                if ((try seen_clients.getOrPut(allocator, cid)).found_existing) continue;
+                if ((try seen_clients.getOrPut(allocator, .{ dev_idx, cid })).found_existing) continue;
             }

             const key = processKey(pid, dev_idx);
```

neudinger:
You are right, I applied your recommendation.

If you think that all, everything look good to me.

catwell:
LGTM too. And thanks for implementing this oneapi support! Looking forward to the next release now :)

neudinger (approved):

PR #262, opened 2025-06-13
build.zig

gwenzek:
```
$ zig build test
[ lots of bazel logs ]
test
└─ run test stderr
debug(zml/async): coro resume CoroId{.cid=0, .i=-1} from CoroId{.cid=0, .i=-1}
warning(zml/tensor): Dim mismatch ! Axis b=2 but received a new tensor where b=3
warning(zml/tensor): Axis not found: c
warning(zml/tensor): Dim mismatch ! Axis d=5 but received a new tensor where d=7
2025-06-13 14:57:03.680394: I xla/service/dump.cc:562] HloModule dump enabled with path prefix: , suffix: before_optimizationsient transposed
2025-06-13 14:57:03.681071: I xla/hlo/utils/hlo_sharding_util.cc:3079] There is no registered layout_canonicalization_callback.
2025-06-13 14:57:03.699419: I xla/hlo/utils/hlo_sharding_util.cc:3079] There is no registered layout_canonicalization_callback.ith mixed integer/t
96/109 ops.test.triton...SKIP
108 passed; 1 skipped; 0 failed.
```

Caveats:
* Require a local bazel installed and tar utility
* Require to have the CPU pjrt plugin visible in the path when running the executables
you can do so eg by `find -L ./bazel-out -name libpjrt_cpu.dylib` then `cp ./bazel-out/.../libpjt_cpu.dylib .`
* I haven't test with the others plugins (probably just a matter of adding flags to build.zig and passing them to bazel)

Issue #67, opened 2024-10-24
Strategy for integrating zml with an existing zig project that is using build.zig


AdjectiveAllison:
Chatted about this in the zml discord. The best current approach seems to be to import a bazel-built zml library into another project on the traditional zig build system. This results in a slightly strange scenario where you end up calling zig from zig but through the cffi layer. It can be workable, but not the most ergonomic. 

This is primarily just an issue to track ideas/progress/feedback on this specific use case. 

It is related, but an adjacent conversation to the "why bazel" one - https://github.com/zml/zml/issues/63

gwenzek:
I have some ideas on how to improve this, 

eg: we could use bazel to generate a `zml.zon` with pointers to build artifacts created by bazel. Then provide a generic build.zig code that call bazel, and use the generated `zml.zon` to use it as a module.

We have upcoming internal deadline, so I don't expect to work on this before December.

PR #651, opened 2026-07-11
docs: explain build.zig integration status

ousamabenyounes:
## Summary

Documents the current path for using ZML from a project that already uses Zig's standard `build.zig` flow. Explains that ZML is Bazel-first today, recommends keeping ZML behind a Bazel-built artifact boundary (C ABI / executable), and notes the future generated-manifest direction discussed in #67.

## Context

As discussed in #67, ZML is currently built with Bazel and there is no native Zig package consumable via `build.zig.zon`. The current recommended path is to build the target with Bazel and consume the resulting artifact from a `build.zig` project via C ABI / executable boundary.

## Prior art

@gwenzek already proposed a manifest-based approach in #67 (Bazel generating a `zml.zon` pointing to build artifacts, consumed by a generic `build.zig` helper), and started an early experiment in #262. That work is still exploratory and not merged, so this PR documents the current status quo rather than the proposed future direction — happy to update the "Future direction" section if there's been more recent progress I've missed.

This is docs-only, no code changes.

## Issues

Refs #67

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,6 +20,7 @@ Ready to write some code? Try starting with [your first model in ZML](./tutorial
 - [Cross Compile and Deploy Models on a Server](./howtos/deploy_on_server.md)
 - [Dockerize Models](./howtos/dockerize_models.md)
 - [Profile ZML Programs](./howtos/profiling.md)
+- [Use ZML from a `build.zig` Project](./howtos/use_from_zig_build.md)
 
 ## Learn more...
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,7 +20,7 @@ Ready to write some code? Try starting with [your first model in ZML](./tutorial
 - [Cross Compile and Deploy Models on a Server](./howtos/deploy_on_server.md)
 - [Dockerize Models](./howtos/dockerize_models.md)
 - [Profile ZML Programs](./howtos/profiling.md)
-- [Use ZML from a `build.zig` Project](./howtos/use_from_zig_build.md)
+- [Use ZML alongside a `build.zig` Project](./howtos/use_from_zig_build.md)
 
 ## Learn more...
 
--- a/docs/howtos/use_from_zig_build.md
+++ b/docs/howtos/use_from_zig_build.md
@@ -1,33 +1,51 @@
-# Use ZML from a `build.zig` project
+# Use ZML alongside a `build.zig` project
 
 ZML is currently built and packaged with Bazel. A native Zig package that can be
 added directly to another project's `build.zig.zon` is not available yet.
 
-If your application already uses Zig's standard build system, keep the ZML build
-as a Bazel step and consume its output from your `build.zig` project. This keeps
-ZML's toolchain, MLIR, PJRT, and platform-specific dependencies under Bazel,
-while your application can stay on the standard Zig build flow.
+If your application already uses Zig's standard build system, keep the ZML
+component behind an explicit artifact boundary. Bazel is the source of truth for
+building ZML and for selecting its toolchain, MLIR, PJRT, and platform-specific
+dependencies.
 
-## Current integration model
+## Current options
 
-1. Build the ZML target you need with Bazel.
-2. Export the produced library or executable artifact from `bazel-bin`.
-3. Link or run that artifact from your Zig project.
-4. Keep the boundary explicit, usually through the C ABI or a command-line
-   executable, instead of trying to import ZML internals as Zig modules.
+The most isolated option is an executable boundary: build and run the ZML
+component with Bazel, then communicate with it through a command-line or
+application-specific interface.
 
-This is less ergonomic than a native Zig package, but it avoids duplicating the
-large Bazel-managed dependency graph in a second build system.
+A `build.zig` project may also consume a Bazel-built library through a C ABI
+facade. In that setup, the application owns the facade, invokes Bazel to build
+the target, and links the produced artifact from the Zig build. This is a
+workaround rather than a packaged integration: ZML does not currently provide a
+maintained C ABI package, generated manifest, or helper for discovering and
+packaging all required runtime dependencies.
+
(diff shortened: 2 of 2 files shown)


gwenzek:
do you have a repo that uses ZML through builld.zig ?

ousamabenyounes:
I do not have an existing public repo that uses ZML this way yet, but I did validate the interim model locally.

I built a small Bazel-side C ABI facade that imports `//zml` and exports two trivial functions, then created a separate `build.zig` consumer outside the ZML repo that links the Bazel-produced shared library and calls those functions.

The local proof passed:

```text
bazel build //repro/cabi:zml_repro_cabi
# produced bazel-bin/repro/cabi/libzml_repro_cabi.so

zig build run
# ZML C ABI consumer OK: KiB=1024, logo_blocks=3
```

So the artifact/C ABI boundary described here is workable. That said, this is still not native `build.zig.zon` integration, and it required an explicit C ABI facade. I can either publish the minimal repro separately, or soften the wording here to make clear this is the current workaround rather than a polished package integration.

gwenzek:
> I built a small Bazel-side C ABI facade that imports //zml and exports two trivial functions, 

can you share your code ?
I'm interested in allowing build.zig to call bazel, but so far I've been struggling to find a way to expose the right binary dependencies.

ousamabenyounes:
It is still a local repro, not a complete integration design, but the external `build.zig` now does call Bazel and discovers the produced artifact through `cquery` before linking it.

The Bazel-side facade is intentionally small:

```python
load("@rules_zig//zig:defs.bzl", "zig_shared_library")

zig_shared_library(
    name = "zml_repro_cabi",
    main = "zml_repro_cabi.zig",
    shared_lib_name = "libzml_repro_cabi.so",
    visibility = ["//visibility:public"],
    deps = ["//zml"],
)
```

The facade imports `zml` and exposes a few pure ZML APIs through C ABI: `KiB`, `MiB`, `logo.isShineGlyph`, `DataType.f32.sizeOf`, `DataType.resolvePeerType`, and `Shape.init(...).byteSize()`.

The external `build.zig` staging step is the more relevant part:

```zig
const stage_zml_artifact = b.addSystemCommand(&.{
    "sh",
    "-c",
    STAGE_BAZEL_ARTIFACT_SCRIPT,
    "stage-zml-artifact",
    bazel,
    ZML_REPRO_TARGET,
    zml_stage_dir,
    zml_root,
    ZML_REPRO_SHARED_LIBRARY,
    zml_header_rel,
});

exe.step.dependOn(&stage_zml_artifact.step);
```

With the shell script doing:

```sh
"$bazel" build "$target"
matches=$("$bazel" cquery --output=files "$target" |
  awk -v name="$library_name" 'BEGIN { count=0 } $0 ~ "/" name "$" { print; count++ } END { if (count != 1) exit 42 }')
artifact="$matches"

cp "$artifact" "$tmp_dir/lib/$library_name"
cp "$header_rel" "$tmp_dir/include/"
test -s "$tmp_dir/lib/$library_name"
test -s "$tmp_dir/include/$(basename "$header_rel")"
mv "$tmp_dir" "$stage_dir"
```

Then the Zig consumer links against the staged directory:

```zig
root_module.addIncludePath(.{ .cwd_relative = zml_include_dir });
root_module.addLibraryPath(.{ .cwd_relative = zml_library_dir });
root_module.linkSystemLibrary("zml_repro_cabi", .{});
root_module.link_libc = true;
```

I also added a `zig build test` step using the same staged library, so it is not only a run smoke test.

Local validation:

```text
zig build run
ZML C ABI consumer OK: KiB=1024, MiB=1048576, logo_blocks=3, f32_size=4, shape_bytes=120

zig build test --summary all
Build Summary: 4/4 steps succeeded; 1/1 tests passed
```

Important limitation: this still does not solve the full binary/runtime dependency problem. It uses an explicit Bazel target and an application-owned C ABI facade. `cquery --output=files` finds the main `.so`, but it is not a generated manifest for transitive runtime dependencies, and this does not exercise a real model/PJRT execution path.

So I think this is useful as a minimal starting point for the `build.zig -> Bazel -> staged artifact -> link/test` shape, but not something I would present as native package integration.

ousamabenyounes:
I published the repro here: https://github.com/ousamabenyounes/zml-bazel-buildzig-repro

I pushed it a bit further than the local snippet I described earlier:

- the recommended path is now Dockerized with `docker run --rm`, so it does not rely on host Zig/Bazel binaries;
- tool inputs are pinned in `repro.lock`: Docker image digest, Zig 0.16.0 tarball URL + SHA256, Bazelisk v1.25.0 URL + SHA256, Bazel 9.1.1;
- `build.zig` installs a tiny Bazel package into a ZML checkout, invokes Bazel, discovers the shared library via `bazel cquery --output=files`, stages the `.so` and header, then links/tests a normal external Zig consumer;
- the shared-library staging now records observed ELF `DT_NEEDED` entries into `needed_libraries.txt` and `zml_repro_manifest.json`, so the repro exposes the direct binary deps it can see from the Bazel-built artifact;
- it also keeps the second executable-boundary path (`run-bazel-exe`) with runfiles/runfiles-manifest staging.

Validated locally:

```text
zig build test -Dzml-root=/home/ousama/www/public/zml -Dbazel=/home/ousama/.cache/bazelisk/downloads/sha256/857bed5d2756b4d998d3caebf2d941d13d434c4eda4b1d6d7dda205736c25a93/bin/bazel --summary all
Build Summary: 4/4 steps succeeded; 1/1 tests passed

scripts/docker-run.sh /home/ousama/www/public/zml test-discovery-failure --summary all
/tmp/zml-repro-tools/zig.tar.xz: OK
/tmp/zml-repro-tools/bazel: OK
strict cquery discovery rejects missing and duplicate artifacts
Build Summary: 2/2 steps succeeded
```

The repo is still intentionally scoped as an artifact/C ABI and executable-boundary repro, not native package integration or a complete runtime dependency manifest generator.

PR #653, opened 2026-07-13
zml: cleanup attention code

gwenzek:


zml/attention/attention.zig
```diff
@@ -151,11 +131,22 @@ pub const Metadata = union(Backend) {
...
+            .metal_fa => |*v| metal.Metadata.deinitBuffer(v),
         }
     }
 };
 
+/// Causal attention as used in transformers.
+///
+/// **Shapes**:
+///   - q, result: .{ .h, .q, .hd }
+///   - k, v:      .{ .h, .k, .hd }
```
Corendos: Are you sure of the shapes ? Isn't it `.{ .q, .h, .hd}` ?

Corendos (approved):
Just curious, when is the `zml$paged_attn` custom call registered on CPU ?

zml/attention/attention.zig
gwenzek: it's actually not very clear, it's not documented. The stablehlo version doesn't care, AFAICT, but flash attention does. Do you know which is it ?

Can I add some asserts here ?

gwenzek:
(sorry the paged_attention code isn't ready , removed)

zml/attention/attention.zig
gwenzek: I confirmed it's indeed (q,h,hd) for the FA2 backend. But there is no explicit transpose / asserts in our code, so it's easy to pass the wrong thing.
the binding also lack proper batching support, and I'm not sure the multi-query per kv head is properly implemented. (the test break when I increase q)

PR #655, opened 2026-07-13
zml: improve Buffer sharding related error messages

gwenzek:
It's easy to break `Buffer.from(x_tensor.shape(), data, sharding)` by creating `x_tensor` in the model code with a specific sharding and you pass a different one to `from`.

I think this should be a crash (like when you mix-up allocators in init/deinit) and it should have a dedicated error message to guide the developers.

### DIFF
--- a/zml/testing.zig
+++ b/zml/testing.zig
@@ -14,9 +14,17 @@ var _platform: ?*const Platform = null;
 pub fn env() *const Platform {
     if (!builtin.is_test) @compileError("Cannot use zml.testing.env outside of a test block");
     if (_platform == null) {
-        _platform = Platform.auto(std.heap.c_allocator, std.testing.io, .{
-            .xla_gpu = .{ .allocator = .{ .bfc = .{ .preallocate = true, .memory_fraction = 0.85 } } },
-        }) catch unreachable;
+        const platform = Platform.auto(
+            std.heap.c_allocator,
+            std.testing.io,
+            .{
+                .xla_gpu = .{ .allocator = .{ .bfc = .{ .preallocate = true, .memory_fraction = 0.85 } } },
+            },
+        ) catch @panic("Pjrt not available");
+
+        // Register shardings required by flash attention.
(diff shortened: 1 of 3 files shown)


Finistere:
Nice errors. IMHO, it would be better to move those log.err inside `axisSplit` where those errors are generated, everyone would benefit from it that way. I would also rather have a `placement(...) catch @panic("<why this cannot/should not fail | broken invariant>")` rather than `placementOrPanic` which provides no information as to why it should just never happen in the places where you changed it.

I'm supposing it's always an irrecoverable error though, which seems to be the case AFAIK

gwenzek:
> I would also rather have a placement(...) catch @panic("<why this cannot/should not fail | broken invariant>")

The error message is supposed to explains why an error at this point is a programming error, and what is the likely root cause. It's on purpose a private helper of `zml.Buffer` because it's aimed to help fix the crash you hit in `zml.Buffer.from`

The same error during compilation of the model would mean something different and would need a different fix.

Finistere (approved):

zml/testing.zig
```diff
@@ -14,9 +14,17 @@ var _platform: ?*const Platform = null;
...
+            std.heap.c_allocator,
+            std.testing.io,
+            .{
+                .xla_gpu = .{ .allocator = .{ .bfc = .{ .preallocate = true, .memory_fraction = 0.85 } } },
+            },
+        ) catch @panic("Pjrt not available");
+
+        // Register shardings required by flash attention.
+        _ = platform.registerSharding("model", .mesh(.{ .model = .high_bandwidth })) catch unreachable;
+        _platform = platform;
```
Corendos, line 27: Not sure of the implications of that because that creates an implicit sharding (but having a global `platform` is also a problem in a way so maybe that's fine)

Corendos (approved):

zml/testing.zig
gwenzek: the problem is that some zml functions relies on specific sharding exist. The test case can't create it, it's too late, because the platform is `const` at this point. Sharing a `platform` instance across test cases is important for the end-to-end duration of the test suite.

PR #656, opened 2026-07-15
zml/tensor: remove `concatenate` and `stack` limits

Corendos:
We can now rely on the compilation context to allocate when needed.

### DIFF
--- a/zml/tensor.zig
+++ b/zml/tensor.zig
@@ -1893,8 +1893,7 @@ pub const Tensor = struct {
     /// Concatenates the input Tensors along the given axis.
     pub fn concatenate(tensors: []const Tensor, axis_: anytype) Tensor {
         if (tensors.len == 1) return tensors[0];
-        stdx.debug.assert(tensors.len <= 32, "concatenate only supports up to 32 tensors, got {}", .{tensors.len});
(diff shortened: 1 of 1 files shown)


gwenzek (approved):

zml/tensor.zig
```diff
@@ -1893,8 +1893,7 @@ pub const Tensor = struct {
     /// Concatenates the input Tensors along the given axis.
     pub fn concatenate(tensors: []const Tensor, axis_: anytype) Tensor {
         if (tensors.len == 1) return tensors[0];
-        stdx.debug.assert(tensors.len <= 32, "concatenate only supports up to 32 tensors, got {}", .{tensors.len});
-        var buffer: [32]*const mlir.Value = undefined;
+        var buffer = CompilationContext.current().arena.allocator().alloc(*const mlir.Value, tensors.len) catch unreachable;
```
gwenzek: we use `panic("OOM")` in other places

Corendos: Addressed in https://github.com/zml/zml/pull/656/commits/9e6c7ad82ac4ca7f0ca9c72d750a1a9536172970

PR #659, opened 2026-07-16
flashattn: harmonize zml.Tensor.zeroes() calls

oboulant:
Following https://github.com/zml/zml/pull/653

### DIFF
--- a/zml/attention/flashattn.zig
+++ b/zml/attention/flashattn.zig
@@ -1006,7 +1006,7 @@ pub const paged_fa2 = struct {
                     .hg = num_head_groups,
                     .b = q.dim(.b),
                 }, .f32)).withPartitioning(.{ .hkv = .model });
-                const dummy_cu_seqlens_k: zml.Tensor = .zeroes(cu_seqlens_q.shape());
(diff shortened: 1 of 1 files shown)


gwenzek:
we kinda prefer the `const x: zml.Tensor = .zeroes` style
https://github.com/zml/zml/blob/master/docs/misc/style_guide.md

oboulant:
Thx for the pointer, I will change it !

oboulant:
Done

gwenzek (commented):
I don't think your code compile. Also I touched those files, so some of the code won't be here anymore. Try rebasing please.

the `.method` pattern doesn't work with the builder pattern that chain methods.
It's fine to keep Tensor first, even if sometimes it's more readable to invert the calls eg `x: Tensor = .add(.scalar(1, f32), .scalar(2,f32))` 

In the current exemple the `withPartitioning(.replicated)` are effectively no-op, so they are better removed imo.

oboulant:
Not much left in the PR

gwenzek (approved):

PR #660, opened 2026-07-16
zml/nn: rework partial rotary factor

Corendos:
This one was tough.

I changed the way rope is computed so hopefully each model can use it without having to reimplement it.

Basically, what changes here is how we apply partial rotary factor. There is a special case for proportional because it doesn't work the same way.

# Context
To give more context, the idea behind partial rotary factor is to rotate only a part of the head. So you have a factor called `partial_rotary_factor` that is used to compute `rotary_dim` using `rotary_dim = partial_rotary_factor * head_dim`. The head will be split into `rotary_dim` and the rest and you'll rotate only the first `rotary_dim` values of the head.

So far, so good.

However, there is a notion of layout in the head. To do rotation, you need to have pairs of values and depending on how the model was trained, pairs are not laid out in the same manner.

In the `interleaved` layout, the head is:
```
| x0 | y0 | x1 | y1 | ... | x_(hd/2) | y_(hd/2) |
```

In the `sequential` layout, it's:
```
| x0 | x1 | ... | x_(hd/2) | y0 | y1 | ... | y_(hd/2) |
```

This is where things get interesting, when the rope scaling is `proportional` we do the pass-through after splitting into `x`s and `y`s. The other scaling do the pass-through before.

# Solution
I struggled a lot because I couldn't understand why non-proportional would do the split before because to me, depending on the layout, it was wrong. When in `interleaved` it's fine because that's identical, but in `proportional`, we get rid of the `y`s.

I guess the idea is that it depends when the layout applies: before of after the split.

In any case, I've implemented a special case for `proportional` because after some search, it seems that it's the only scaling that do the proportional thingy after the split.

# Additional details
What we used to do before was that we would set the frequencies returned by `invFreq` to 0 after `rotary_dim`. This is identical to the current solution but we couldn't apply the same trick to non-`proportional` scaling.

### DIFF
--- a/zml/nn.zig
+++ b/zml/nn.zig
@@ -182,6 +182,7 @@ pub const RopeOpts = struct {
             rope_theta: f32 = 10000,
             llama_4_scaling_beta: ?f32 = null,
             attention_factor: ?f32 = null,
+            partial_rotary_factor: f32 = 1.0,
         };
 
         pub const Linear = struct {
@@ -190,7 +191,7 @@ pub const RopeOpts = struct {
         };
 
         pub const Proportional = struct {
-            partial_rotary_factor: f32,
+            partial_rotary_factor: f32 = 1.0,
             rope_theta: f32 = 10000,
         };
 
@@ -265,6 +266,13 @@ pub const RopeOpts = struct {
                 inline else => |s| s.rope_theta,
             };
         }
+
+        pub fn getPartialRotaryFactor(self: *const Scaling) f32 {
+            return switch (self.*) {
+                inline .yarn, .proportional => |s| s.partial_rotary_factor,
+                .default, .llama3, .linear => 1.0,
+            };
+        }
     };
 };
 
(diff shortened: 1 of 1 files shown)


zml/nn.zig
```diff
@@ -287,11 +296,26 @@ pub fn rope(x: Tensor, pos_idx: ?Tensor, opts: RopeOpts) Tensor {
         stdx.debug.assert(x.shape().hasTags(.{ .s, .hd }), "rope expects x argument to have both .s and .hd axes got: rope(x={f})", .{x});
         break :blk Tensor.arange(.{ .end = x.dim(.s) }, .f32).withTags(.{.s});
     };
-    const x_real, const x_imag = splitRealImg(x, opts.layout);
+
+    // NOTE(Corentin): Don't ask
```
tgautie, line 300: 😅

gwenzek (commented):
it would help if you could write the math formula.

Not fan of the change,
I'd rather have a new rope: `zml.nn.ropeProportional` that does it's own thing.
And call it right at the top of rope instead of muddying the water for the simple rope.

This idea of `zml.nn.rope` is to move as much logic as possible in the `invFreq` to simplify the runtime code.

zml/nn.zig
```diff
@@ -287,11 +296,26 @@ pub fn rope(x: Tensor, pos_idx: ?Tensor, opts: RopeOpts) Tensor {
...
+    // Well, okay if you insist. It seems that the proportional rope doesn't really work like the others with respect to partial rotary factor.
+    // Instead of splitting x into `rot` and `pass` it first splits into real and imag (thus respecting the layout) and then rotates only a part.
+    // If you ask me, that's the correct way of doing partial rotation semantically but ¯\_(ツ)_/¯.
+    if (opts.scaling == .proportional) {
+        return proportionalRope(x, idx, opts);
+    }
+
+    // NOTE(Corentin): defaulting to 1.0 makes it the degenerate case that is identical to no
```
tgautie, line 308: Unfinished comment: "identical to no partial rotation"?

Corendos:
Closing in favor of #662

PR #662, opened 2026-07-16
zml: handle passthroug in zml.nn.rope

gwenzek:
alternative to #660


Corendos (approved):
I think that's better

PR #669, opened 2026-07-20
zml: ground work to support Poolside's Laguna models

Corendos:
This PR contains two main changes:
* A patch for the tokenizer to support the tokenizer config file
* A temporary workaround to speed-up loading of the model when executing arbitrary executable at load time

### DIFF
--- a/zml/io.zig
+++ b/zml/io.zig
@@ -447,8 +447,9 @@ pub const Loader = struct {
         defer for (buffers, loaded) |*b, l| if (l) b.deinit();
 
         for (sources, 0..) |source, i| {
-            self.loadSingle(io, source, source.shape, &buffers[i], &loaded[i], shardings, .{});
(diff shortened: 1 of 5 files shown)


Finistere (approved):

zml/io.zig
```diff
@@ -447,8 +447,9 @@ pub const Loader = struct {
         defer for (buffers, loaded) |*b, l| if (l) b.deinit();
 
         for (sources, 0..) |source, i| {
-            self.loadSingle(io, source, source.shape, &buffers[i], &loaded[i], shardings, .{});
+            self.group.async(io, loadSingle, .{ self, io, source, source.shape, &buffers[i], &loaded[i], shardings, .{} });
         }
+        try self.group.await(io);
```
Corendos, line 452: This will need to be reverted once we have a better solution for io

PR #674, opened 2026-07-21
Support quantization (Metal only for now)

elogir:


zml/ops.zig
```diff
@@ -1514,6 +1514,101 @@ pub fn customCallOutputOperandAliases(
...
+    for (inputs, 0..) |t, i| {
+        operand_values[i] = t.value();
+    }
+
+    const result_types = allocator.alloc(*const mlir.Type, outputs.len) catch @panic("OOM");
+    for (outputs, 0..) |s, i| {
+        result_types[i] = mlirx.Type.rankedTensor(mlir_ctx, s);
+    }
+
+    const op = mlir.Operation.make(mlir_ctx, "stablehlo.composite", .{
```
gwenzek, line 1591: can you add dedicated bindings in stablehloo.zig ?

examples/llm/models/qwen3_5_moe/model.zig
```diff
@@ -757,6 +757,8 @@ pub const Moe = struct {
             self.down_proj,
             null,
             null,
+            null,
```
gwenzek, line 760: at this point it would be better to have a struct representing forwardMoe arguments.

zml/platform.zig
gwenzek, line 657: steeve wanted us to remove `xla_gpu: XlaGpu` and use `cuda: XlaGpu`, `rocm: XlaGpu` instead.

zml/platform.zig
```diff
@@ -658,7 +658,7 @@ pub const CreateOptions = struct {
     tpu: struct {} = .{},
     neuron: struct {} = .{},
     oneapi: struct {} = .{},
```
gwenzek, line 660: IIUC the switch below `oneapi` should also have the type `XlaGpu`

zml/platform.zig
```diff
@@ -658,7 +658,7 @@ pub const CreateOptions = struct {
     tpu: struct {} = .{},
```
gwenzek, line 658: what do we gain by having `struct {}` instead of `void` ?

zml/platform.zig
```diff
@@ -727,7 +736,9 @@ pub const CreateOptions = struct {
         values.shrinkRetainingCapacity(0);
         switch (target) {
             .cpu => self.cpu.writeNamedValues(&values),
-            .cuda, .rocm, .oneapi, .metal => self.xla_gpu.writeNamedValues(target, &values),
+            .cuda, .rocm, .oneapi => self.xla_gpu.writeNamedValues(&values),
+            .metal => self.metal.writeNamedValues(&values),
```
gwenzek, line 740: this should be one branch:

`inline .cpu, ..., .metal => |opts| opts.writeNamedValues(&values)`

zml/ops.zig
```diff
@@ -1514,6 +1514,208 @@ pub fn customCallOutputOperandAliases(
...
+    const outs = composite("xla.scaled_dot", &.{ lhs, rhs, lhs_scale, rhs_scale }, &.{res_shape}, scaledDotReference, res_shape, .{
+        .composite_attributes = &.{.named(mlir_ctx, "dimension_numbers", dnums)},
+    });
+
+    return outs[0];
+}
+
+fn scaledDotReference(in: []const Tensor, out_shape: Shape) Tensor {
+    return customCall(
+        "zml$scaled_dot_unmatched",
```
gwenzek, line 1711: what's this custom call ?

zml/ops.zig
elogir: Will do!

zml/ops.zig
elogir: It's temporary. It's not implemented on the XLA side, it's just a way to make it panic for debug purposes in case of an unmatched composite that gets decomposed. What's actually needed is a stablehlo or something fallback in case it's not handled in XLA

PR #679, opened 2026-07-22
zml/testing: Do not pre-allocate memory for the zml.testing.env()

Finistere:
Pre-allocating 85% of memory for simple tests is overkill, we'll just let the BFC grow as needed. That should hopefully limit the problem of having concurrent CI runs.


oboulant (approved):
LGTM

PR #680, opened 2026-07-22
//zml/io/vfs: promote to //vfs

gwenzek:
Proposals:

1. simplify vfs architecture: merge `//zml/io` and `//zml/io/vfs`. They have the same deps, and we only use `//zml/io`. New name: `//zml/vfs` 

2. move vfs to top level (it becomes `//vfs`)

3. ~~expose vfs and stdx to any zig user through `zig fetch` by adding `build.zig` and `build.zig.zon`~~ abandonned

All of the above are no-op for current consumer of zml, they can still access vfs through `zml.io.VFS`

The PR also include two bug fix:
- S3: when no content length we were trying to read exactly 1M, while we should read up to 1M
- GCP: restore the commented code for parsing bucket info

## Test plan:

```
bazel run --config=debug //examples/io:playground -- ls gs://gcp-public-data-landsat/
bazel run --config=debug //examples/io:playground -- tree s3://noaa-goes19/ABI-Flood-Day-Shapefiles/2025/08

zig build run_vfs_example -- tree s3://noaa-goes19/ABI-Flood-Day-Shapefiles/2025/08
zig build run_vfs_example -- ls gs://gcp-public-data-landsat/
```


vfs/example.zig
gwenzek, line 1: this is a fork of `examples/io/main.zig` that removes all the tensor related stuff, and therefore has no deps beyond vfs.
I don't know if it's worth maintaining the two of them.

examples/io/main.zig
```diff
@@ -72,15 +73,15 @@ pub fn main(init: std.process.Init) !void {
 
             try stdout_writer.interface.print("Wrote {B:.2} to stdout from {s}\n", .{ read, path });
         },
-        .ls => {
+        .ls, .tree => {
```
gwenzek, line 76: the default `ls` would take forever if you give it a huge bucket because of the recursive nature of `printTree`. I've make `ls` more similar to the Posix version by being non-recursive, and `tree` keeps `ls` previous behavior.

vfs/example.zig
Finistere: I'm not sure to see what the benefit is, seems like just extra code. I think it's fine to show an example with zml, it's not that remove to take only the parts one is interested in. I would expect this file to just become abandoned over time.

Finistere (approved):

gwenzek (commented):
Merging without build.zig and vfs/example.zig

PR #681, opened 2026-07-24
zml/attention: test paged attention

gwenzek:
Add unit test for paged attention, enforcing all our backends produce the same outputs.

To simplify memory management of many small buffer provided in a struct (like paged attention metadata buffers), I introduce `zml.Buffer.deinitAll` backed by `zml.meta.visitFlatStruct`. Contrary to og `zml.meta.visit` the `flatStruct` variant emit a  compile error if the input has a slice/ptr inside it. The idea is that since it enforces the struct to be a trivial, it means the struct won't require complex `deinit` and `zml.Buffer.deinitAll` is in fact everything you need to `deinit`. This avoid previous footgun of `zml.io.unloadAll`.

This led to more improvement with `zml.Tensor.toMemoryAll` and `zml.Tensor.onMemoryAll` that also leverage the new `zml.meta.visitFlatStruct`

### DIFF
--- a/zml/attention/paged_attention.zig
+++ b/zml/attention/paged_attention.zig
@@ -272,3 +272,143 @@ test "Backend.auto selects triton on oneAPI" {
 
     try std.testing.expectEqual(Backend.triton, Backend.auto(&platform));
 }
+
+test pagedAttention {
+    const platform = zml.testing.env();
+    const io = std.testing.io;
+    const allocator = std.testing.allocator;
+    var arena_state: std.heap.ArenaAllocator = .init(allocator);
+    defer arena_state.deinit();
+    const arena = arena_state.allocator();
+
+    const batch_size = 8;
+    const num_pages = 64;
+    const page_size = 16;
+    const tensors: struct { q: zml.Tensor, k: zml.Tensor, v: zml.Tensor, kv_cache: KvCache, token_index: zml.Tensor } = .{
+        .q = .init(.{ .b = 8, .hkv = 2, .hg = 4, .hd = 32 }, .f32),
+        .k = .init(.{ .b = 8, .hkv = 2, .k = 16, .hd = 32 }, .f32),
+        .v = .init(.{ .b = 8, .hkv = 2, .k = 16, .hd = 32 }, .f32),
+        .kv_cache = .{
+            .split = .{
+                .k = .init(.{ .page = num_pages, .hkv = 2, .k_chunk = page_size, .hd = 32 }, .f32),
+                .v = .init(.{ .page = num_pages, .hkv = 2, .k_chunk = page_size, .hd = 32 }, .f32),
+            },
+        },
+        .token_index = .init(.{}, .u32),
+    };
+
+    const rng_q = try platform.compileFn(allocator, io, zml.Tensor.Rng.normal, .{ tensors.q.shape(), .{} }, .{});
+    defer rng_q.deinit();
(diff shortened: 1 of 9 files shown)


zml/attention/paged_attention.zig
```diff
@@ -24,7 +25,24 @@ pub const Backend = enum {
             .oneapi => .triton,
             .tpu => .mosaic_tpu,
             .metal => .metal,
-            else => stdx.debug.panic("Paged attention is not supported on {s} yet", .{@tagName(platform.target)}),
+            // .cpu => .vanilla
+            .cpu, .neuron => stdx.debug.panic("Paged attention is not supported on {s} yet", .{@tagName(platform.target)}),
+        };
+    }
+
+    pub fn isAvailable(backend: Backend, platform: *const zml.Platform) bool {
```
Finistere, line 39: nice

zml/attention/paged_attention.zig
```diff
@@ -215,6 +209,44 @@ pub const KvCache = union(enum) {
...
+
+    pub fn update(
+        self: KvCache,
+        new_k: zml.Tensor,
+        new_v: zml.Tensor,
+        slot_mapping: zml.Tensor,
+        chunk_size: u32,
+        backend: Backend,
+    ) KvCache {
+        const active_page, const k_chunk = getPageAndOffsetFromSlotMapping(slot_mapping, chunk_size);
```
Finistere: Just a detail, maybe keep `page_index` and `offset` from the `getPageAndOffsetFromSlotMapping` function? Feels a bit clearer than `active_page` and `k_chunk` IMHO

zml/module.zig
```diff
@@ -170,6 +170,10 @@ pub const CompilationContext = struct {
         _current = null;
     }
 
+    pub fn isActive() ?*CompilationContext {
```
Finistere: Weird naming IMHO, I would expect a boolean here. I would use `currentOrNull` given the current structure.

zml/meta.zig
```diff
@@ -606,6 +606,56 @@ test visit {
     }
 }
 
+/// Similar to `visit` but doesn't follow pointers or slices.
+/// This guarantees that the given type is a flat struct.
+///
+/// see: `zml.meta.visit`
+pub fn visitStruct(comptime callback: anytype, ctx: FnParam(callback, 0), v: anytype) VisitReturn(callback) {
```
Finistere: You're using `visit` within this function, so only the top level struct needs to be flat here. I would consider naming this `visitFlatStruct`

zml/tensor.zig
```diff
@@ -148,10 +148,22 @@ pub const Tensor = struct {
...
         const partitioned_shape = self._shape.withPartitioning(axes_);
 
-        const attr = ctx.partitioning.tensorShardingAttr(ctx.allocator, ctx.mlir_ctx, partitioned_shape, null) catch unreachable;
+        const ctx = CompilationContext.isActive() orelse {
+            var res = self;
+            res._shape = partitioned_shape;
+            return res;
+        };
+
+        const attr = ctx.partitioning.tensorShardingAttr(ctx.allocator, ctx.mlir_ctx, partitioned_shape, null) catch |err| switch (err) {
```
Finistere, line 159: nice

Finistere (approved):

zml/meta.zig
gwenzek: oops

PR #682, opened 2026-07-28
zml: simplify memory allocation during compilation

gwenzek:
My original proposition was to use setjmp/longjmp to properly recover from OOM during compilation.

We haven't reach an agreement on wether it's a good idea or not, so I removed that part,
and kept the part that simplify memory allocation when compiling.

### DIFF
--- a/zml/module.zig
+++ b/zml/module.zig
@@ -198,10 +198,13 @@ pub const CompilationContext = struct {
         return self.channel_id;
     }
 
-    pub fn abortOOM() noreturn {
-        const self = CompilationContext.current();
+    pub fn abortOOM(self: *CompilationContext) noreturn {
         std.debug.assert(self.oom_jmp_buf_active);
         c.longjmp(&self.oom_jmp_buf, 1);
+    }
+
+    pub fn alloc(self: *CompilationContext, T: type, n: usize) []T {
+        return self.arena.allocator().alloc(T, n) catch self.abortOOM();
     }
 };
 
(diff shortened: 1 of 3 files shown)


zml/module.zig
```diff
@@ -232,14 +249,27 @@ pub fn compile(
...
+    defer allocator.destroy(compilation_context);
+
+    compilation_context.* = .init(allocator, st_io.io(), platform, opts);
     defer compilation_context.deinit();
 
-    const result = emitMlir(&compilation_context, func, args) catch unreachable;
+    compilation_context.oom_jmp_buf_active = true;
+    defer compilation_context.oom_jmp_buf_active = false;
+
+    if (c.setjmp(&compilation_context.oom_jmp_buf) != 0) {
```
Finistere: The problem here is that all the defer/errdefer won't be executed, not only for zml's code but also user code that defines the model which is arbitrary. Unless I'm missing something, I don't see how that's safe

gwenzek: which `defer` ? what could be in those `defer` that would break safety ?

Finistere: freeing allocation, closing files, unlocking a mutex, etc. Unless I miss something, anywhere a user calls something like `Tensor.concatenate` it might jump back to the compile call. That's a really weird and unexpected flow.

If you really want to remove the panics, maybe the compile context can allocate once a big enough buffer? I've no idea for what the allocations are used though, works for temporary stuff, more tricky to get right if they're permanent.

gwenzek: all allocations are from the CompilationContext arena, which is correctly freed by the cleanup fonction.
we don't open files while generating the IR, and we don't use mutex either. I'm not sure what use case you have in mind.

Finistere (approved):

PR #683, opened 2026-07-28
Gw/llama walsh

gwenzek:


sboulmier (approved):
Seems good to me, my main remark is that we usually reason in terms of the walsh hadamard transform rather than the hadamard matrix (https://en.wikipedia.org/wiki/Hadamard_transform). The reason is that the WHT is a rotation (it's just the scaled hadamard matrix) so WHT(WHT(x)) = x. Moreover the fast WHT doesn't materialize H, it computes the product with a few in place vector operations, which can matter for large dimensions. One caveat is that a Hadamard matrix can eat and remember the sign randomization, while a FWHT needs to apply signs beforehand in forward mode, and afterwards in reverse mode, so more bookkeeping is needed.

gwenzek:
the goal wasn't to merge, just to share experiment.
For Q and K WHT is a very small matmul, 128x128 I don't expect it to be a bottleneck anywhere. For other parts of the model it can be moved to weight loading and is therefore virtually free.

PR #685, opened 2026-07-29
zml/Slice: correctly handle strides

gwenzek:
While working on paged attention I realize that `zml.testing.expectClose` wasn't properly handling strided Slice.
To fix that I'm adding a `ContiguousItemsIterator` that loops over the contiguous part of the slice.
Only used in test now, so no perf considerations.

### DIFF
--- a/zml/slice.zig
+++ b/zml/slice.zig
@@ -135,11 +135,15 @@ pub const Slice = struct {
 
     pub fn data(slice: Slice) []u8 {
         std.debug.assert(slice.mutable);
-        return @constCast(slice.bytes[slice.offset_bytes..]);
+        std.debug.assert(slice.isContiguous());
+        const end = slice.offset_bytes + slice.shape.byteSize();
+        return @constCast(slice.bytes[slice.offset_bytes..end]);
     }
 
     pub fn constData(slice: Slice) []const u8 {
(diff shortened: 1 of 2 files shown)


zml/slice.zig
```diff
@@ -150,6 +154,87 @@ pub const Slice = struct {
...
+                var contiguous_axis_start: u4 = rank;
+                var contiguous_item_count: usize = 1;
+                var expected_stride: i64 = @intCast(@sizeOf(T));
+
+                var axis = rank;
+                while (axis > 0) {
+                    axis -= 1;
+                    const dim = slice.shape.dim(axis);
+                    const stride = slice.byte_strides.get(axis);
+                    if (dim != 1 and stride != expected_stride) break;
```
Finistere, line 186: I'm a bit confused by the `slice.byte_strides`, why do we need to do this computation if we have `byte_strides` already? why the equality check with `expected_stride`?

Finistere (approved):

zml/slice.zig
gwenzek: this code is reading `slice.byte_strides` from right to left, and compute how many axes are contiguous. If axis == rank, it means the tensor is fully contiguous, and it can be described with one slice.

PR #686, opened 2026-07-29
zml.ops: refactor while loop API

gwenzek:
removes `anytype` from `zml.ops.while`, helping with type checking and simplifying a bit the logic.
A `While` struct needs 
* a State struct definition that captures all the loop state that will change on each iteration
* a `cond(While, State) Tensor` function returning a scalar boolean tensor
* a `body(While, State) State` function that preserves the state shapes.

example:

```
const While = struct {
    // While struct can contain tensor and non-tensor fields.
    limit: i64,
    step: Tensor,

    pub const State = struct {
        i: zml.Tensor,
        sum: zml.Tensor,
    };

    pub fn cond(self: @This(), state: State) Tensor {
        return state.i.cmp(.LT, Tensor.scalar(self.limit, .i64));
    }

    pub fn body(self: @This(), state: State) State {
        return .{
            .i = state.i.add(self.step),
            .sum = state.sum.add(self.step),
        };
    }

    fn forward(i: Tensor, sum: Tensor) State {
        return zml.ops.@"while"(
            @This(),
            .{ .limit = 10, .step = .scalar(1, .i64) },
            .{ .i = i, .sum = sum },
        );
    }
};
```

The code is a bit more verbose for simple example, but for GatedDeltaNet it's equivalent, and I have also used it for paged attention.

### DIFF
--- a/zml/nn.zig
+++ b/zml/nn.zig
@@ -1356,80 +1356,92 @@ pub fn sdpa(q_: Tensor, k_: Tensor, v_: Tensor, opts: SdpaOpts) Tensor {
 }
 
 pub const GatedDeltaNet = struct {
+    /// Query tensor .{ .s, .h, .k }.
+    queries: Tensor,
+    /// Key tensor .{ .s, .h, .k }.
+    keys: Tensor,
+    /// Value tensor .{ .s, .h, .v }.
+    values: Tensor,
+    /// Forget gate .{ .s, .h }.
+    alphas: Tensor,
+    /// Delta gate .{ .s, .h }.
+    betas: Tensor,
+
     pub const State = struct {
         /// Per-head recurrent state with shape .{ .h, .v, .k }.
         s: Tensor,
+        /// Sequence of outputs with shape .{ .s, .h, .v }.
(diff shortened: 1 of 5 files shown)


zml/nn.zig
```diff
@@ -1356,83 +1356,95 @@ pub fn sdpa(q_: Tensor, k_: Tensor, v_: Tensor, opts: SdpaOpts) Tensor {
...
+            .step = state.step.addConstant(1),
+        };
+    }
+
     fn sliceStep(input: Tensor, step_: Tensor) Tensor {
         return input.dynamicSlice(.{ .s = Tensor.DynSlice{ .start = step_, .len = 1 } }).squeeze(.s);
     }
 
-    fn validateStep(state: State, inputs: StepInputs) void {
+    fn validateStep(gdn: GatedDeltaNet, state: State) void {
```
tgautie: Should it be validateInitialState now that it does not validate each step but only the input once?

tgautie (approved):
Super nice to have named outputs!

hugomano (approved):

PR #688, opened 2026-07-29
zml/attention: stablehlo paged attention

gwenzek:
Implement paged_attention in pure stablehlo.

introduces `zml.ops.if` and `zml.ops.if2`
The first one is similar to `zml.ops.while` and requires a full struct with two functions. it's a bit heavy to use, but it follows the stablehlo API, and if one of the branch has side effects it will correctly evaluate the side effect only in that branch.

`zml.ops.if2` is a simpler version that doesn't require to create intermediary struct, but as a consequence, if you have side effect they will always evaluate, and it relies on the compiler to correctly avoid duplicated compute by moving relevant compute inside the branches.

(open to feedback  about if/if2)

### DIFF
--- a/zml/attention/paged_attention.zig
+++ b/zml/attention/paged_attention.zig
@@ -598,7 +598,7 @@ const AttentionLoop = struct {
                 page_size: u32,
 
                 /// Prefill version: num_slots == page_size, we multiply a full page of queries with a page of keys.
-                pub fn ifTrue(if_ctx: IfPrefill) State {
+                pub fn onTrue(if_ctx: IfPrefill) State {
                     const while_body = if_ctx.while_body;
                     const state = if_ctx.state;
                     const page_size = if_ctx.page_size;
@@ -619,7 +619,7 @@ const AttentionLoop = struct {
                 }
 
                 /// Decode version: num_slots == 1, we multiply one query with a page of keys.
-                pub fn ifFalse(if_ctx: IfPrefill) State {
+                pub fn onFalse(if_ctx: IfPrefill) State {
                     const while_body = if_ctx.while_body;
                     const state = if_ctx.state;
                     const page_size = if_ctx.page_size;
--- a/zml/attention/paged_attention.zig
+++ b/zml/attention/paged_attention.zig
@@ -22,7 +22,7 @@ pub const Backend = enum {
     triton,
     mosaic_tpu,
     metal,
-    // vanilla,
+    stablehlo,
 
     pub fn auto(platform: *const zml.Platform) Backend {
         return switch (platform.target) {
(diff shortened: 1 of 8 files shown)


zml/attention/paged_attention.zig
```diff
@@ -513,3 +529,247 @@ test pagedAttention {
...
+        var partial = partialSoftmax(attn_weights.convert(precise_dt), .k_chunk);
+        partial.values = partial.values.dot(active_v.convert(precise_dt), .k_chunk).transpose(og_softmax.values.shape());
+        partial.max_value = partial.max_value.transpose(og_softmax.max_value.shape());
+        partial.exp_sum = partial.exp_sum.transpose(og_softmax.exp_sum.shape());
+
+        return og_softmax.merge(partial);
+    }
+
+    pub fn updateState(state: State, num_slots: u32, query_has_more_keys: zml.Tensor, sequence_has_more_queries: zml.Tensor, softmax: PartialSoftmax) State {
+        return zml.ops.if2(
```
gwenzek, line 688: I think this illustrate well we might want to version of `if`. 
This `if` is quite simple and there are no side effect on either branch, so the compiler will trivially be able to move the `dynamicUpdateSlice` to the second branch (in the MLIR it's actually outside of the if/else blocks)

Here introducing another struct to capture all the inputs, would be quite cumbersome.

zml/attention/paged_attention.zig
```diff
@@ -31,14 +31,14 @@ pub const Backend = enum {
             .oneapi => .triton,
             .tpu => .mosaic_tpu,
             .metal => .metal,
-            // .cpu => .vanilla
-            .cpu, .neuron => stdx.debug.panic("Paged attention is not supported on {s} yet", .{@tagName(platform.target)}),
+            .cpu => .stablehlo,
+            .neuron => stdx.debug.panic("Paged attention is not supported on {s} yet", .{@tagName(platform.target)}),
```
gwenzek, line 35: @hugomano do you want to test this on neuron ?

zml/attention/paged_attention.zig
```diff
@@ -513,3 +529,247 @@ test pagedAttention {
...
+            .cmp(seq_num_queries_, .GT, .scalar(1, .i32)),
+        );
+    }
+
+    pub fn attentionOnePage(self: While, state: State, q_offset: zml.Tensor, q_chunk: u32, page_size: u32) PartialSoftmax {
+        const k_offset = state.k_page_idx.scale(page_size);
+        const active_q = self.q.dynamicSlice1d(self.q.axis(.b), .{ .start = state.slot_id, .len = q_chunk });
+
+        const active_k_page_id = self.parameters.block_table.dynamicSlice(.{
+            .b = zml.Tensor.DynSlice{ .start = state.seq_id, .len = 1 },
```
gwenzek, line 655: I find this really horrid, I have a branch with better slice syntax, will propose it.

zml/attention/paged_attention.zig
```diff
@@ -513,3 +529,247 @@ test pagedAttention {
...
+        return zml.ops.@"if"(
+            struct {
+                const IfPrefill = @This();
+                while_body: While,
+                state: State,
+                seq_num_queries: zml.Tensor,
+                page_size: u32,
+
+                /// Prefill version: num_slots == page_size, we multiply a full page of queries with a page of keys.
+                pub fn ifTrue(if_ctx: IfPrefill) State {
```
Finistere: Not for this PR, but maybe `fn onTrue` and `fn onFlase` might be clearer. Seems a bit confusing to have `IfPrefill.ifTrue`

Finistere (approved):

zml/attention/paged_attention.zig
gwenzek: donw

PR #690, opened 2026-07-30
zml: start of quantization support

elogir:
Add MLIR composites, `scaledDot`, and a Linear path for NVFP4 weights (unpack, optional activation quant on Blackwell, global scales). Skip fused multi-source tensors in the generic loader so pack/`loadExecute` can fill them (may be used later for fused QKV etc).

### DIFF
--- a/zml/io.zig
+++ b/zml/io.zig
@@ -373,7 +373,8 @@ pub const Loader = struct {
         };
 
         if (sources.len != 1) {
-            std.debug.panic("Expected loaded tensor to have only 1 source, got {}", .{sources.len});
+            log.debug("Skipping fused tensor with {} sources; expected to be loaded via loadExecute", .{sources.len});
+            return;
         }
 
         self.loadSingleInner(io, sources[0], tensor.shape(), buffer, shardings, opts) catch |e| {
--- a/zml/nn.zig
+++ b/zml/nn.zig
@@ -17,6 +17,9 @@ pub const Linear = struct {
     weight: Tensor,
     bias: ?Tensor = null,
     tag: Shape.Tag,
+    scales: ?Tensor = null,
+    global_scale: ?Tensor = null,
+    input_global_scale: ?Tensor = null,
 
     pub fn init(weight: Tensor, bias: ?Tensor, tag: anytype) Linear {
         return .{
@@ -26,12 +29,53 @@ pub const Linear = struct {
         };
     }
 
-    pub fn forward(self: Linear, x: Tensor) Tensor {
-        var y = x.dot(self.weight, self.tag);
+    pub fn unloadBuffers(self: *zml.Bufferized(Linear)) void {
+        self.weight.deinit();
+        if (self.bias) |*bias| bias.deinit();
+        if (self.scales) |*scales| scales.deinit();
+        if (self.global_scale) |*gs| gs.deinit();
+        if (self.input_global_scale) |*igs| igs.deinit();
+    }
 
+    pub fn forward(self: Linear, x: Tensor) Tensor {
+        const y = self.forwardWeight(x);
         return if (self.bias) |bias| y.add(bias.broad(y.shape())) else y;
     }
+
+    fn forwardWeight(self: Linear, x: Tensor) Tensor {
+        const scales = self.scales orelse return x.dot(self.weight, self.tag);
+
+        const weight = if (self.weight.dtype() == .u8)
+            ops.unpackNvfp4(self.weight.withTags(.{ .dout, .kw }), self.tag)
+        else
+            self.weight;
+
+        const scalar_f32 = zml.Shape.init(.{}, .f32);
+        const igs: ?Tensor = if (self.input_global_scale) |g| g.convert(.f32).reshape(scalar_f32) else null;
+        const wgs: ?Tensor = if (self.global_scale) |g| g.convert(.f32).reshape(scalar_f32) else null;
+
+        const platform = zml.module.CompilationContext.current().platform;
+        if (weight.dtype() == .f4e2m1 and platform.target == .cuda) {
+            const q_packed, const q_scale = ops.quantizeNvfp4(x.convert(.bf16), igs, self.tag);
+            const q = ops.unpackNvfp4(q_packed, self.tag);
+            const acc = ops.scaledDot(q, weight, q_scale, scales, self.tag);
+            return applyGlobalScale(acc, igs, wgs).convert(x.dtype());
+        }
+
+        const acc = ops.scaledDot(x.convert(.bf16), weight, null, scales, self.tag);
+        return applyGlobalScale(acc, null, wgs).convert(x.dtype());
+    }
 };
+
+fn applyGlobalScale(acc: Tensor, igs: ?Tensor, wgs: ?Tensor) Tensor {
+    const combined: ?Tensor = if (igs) |i|
+        (if (wgs) |w| i.mul(w) else i)
+    else
+        wgs;
+    const alpha = combined orelse return acc;
+    const f32_acc = acc.convert(.f32);
+    return f32_acc.div(alpha.broad(f32_acc.shape())).convert(acc.dtype());
+}
 
 pub const TokenEmbedding = struct {
     weight: Tensor,
--- a/zml/nn.zig
+++ b/zml/nn.zig
@@ -55,7 +55,7 @@ pub const Linear = struct {
         const wgs: ?Tensor = if (self.global_scale) |g| g.convert(.f32).reshape(scalar_f32) else null;
 
         const platform = zml.module.CompilationContext.current().platform;
-        if (weight.dtype() == .f4e2m1 and platform.target == .cuda) {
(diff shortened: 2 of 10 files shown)


zml/nn.zig
```diff
@@ -26,13 +29,64 @@ pub const Linear = struct {
         };
     }
 
-    pub fn forward(self: Linear, x: Tensor) Tensor {
-        var y = x.dot(self.weight, self.tag);
+    pub fn unloadBuffers(self: *zml.Bufferized(Linear)) void {
+        self.weight.deinit();
```
gwenzek: note: I've added `zml.Buffer.deinitAll(Linear, self)`

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+    opts: CompositeOpts,
+) []Tensor {
+    const ctx = CompilationContext.current();
+    const mlir_ctx = ctx.mlir_ctx;
+    const allocator = ctx.arena.allocator();
+
+    const decomp_name = std.fmt.allocPrint(allocator, "{s}.impl_{d}", .{ name, ctx.nextCompositeId() }) catch @panic("OOM");
+
+    {
+        const block_types = allocator.alloc(*const mlir.Type, inputs.len) catch @panic("OOM");
```
gwenzek: note: you can use `ctx.alloc(...)` directly now, and you can drop the `panic(oom)`

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+        out_tensors[i] = Tensor._result(s, op.result(i));
+    }
+
+    return out_tensors;
+}
+
+pub fn unpackNvfp4(w: Tensor, k_tag: anytype) Tensor {
+    stdx.debug.assert(w.dtype() == .u8, "unpackNvfp4 expects packed u8 weights, got {}", .{w.dtype()});
+    return w.bitCast(.f4e2m1)
+        .merge(.{ .kb = .{ .kw, .bitcast } })
```
gwenzek: what's `.kw` ? note that you're probably the first one to use `bitCast` seriously. If it's not convenient to have `bitCast` inserting an axis, you can change that directly in `bitCast` (currently it follows stablehlo that insert an axis)

Would it be simpler to use the same axis tag for input / output ?

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+        .renameTag(.kb, Shape.toTag(k_tag));
+}
+
+pub fn quantizeNvfp4(x: Tensor, input_global_scale: ?Tensor, args: anytype) struct {
+    values: Tensor,
+    scales: Tensor,
+} {
+    const block_size = 16;
```
gwenzek: can you pass `block_size` as an argument ?

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+}
+
+pub fn unpackNvfp4(w: Tensor, k_tag: anytype) Tensor {
+    stdx.debug.assert(w.dtype() == .u8, "unpackNvfp4 expects packed u8 weights, got {}", .{w.dtype()});
+    return w.bitCast(.f4e2m1)
+        .merge(.{ .kb = .{ .kw, .bitcast } })
+        .renameTag(.kb, Shape.toTag(k_tag));
+}
+
+pub fn quantizeNvfp4(x: Tensor, input_global_scale: ?Tensor, args: anytype) struct {
```
gwenzek: can you rename `args` to `axis` or maybe `block_axis` ?

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+        stdx.debug.assert(lhs._shape.dim(l) == rhs._shape.dim(r), "scaledDot expects batching dimensions to be equal, got {} and {} in {f} and {f}", .{ l, r, lhs, rhs });
+        var t = lhs._shape.tag(l);
+        if (t == Shape.TagUnknown) t = rhs._shape.tag(r);
+        res_shape = res_shape.appendDim(lhs._shape.dim(l), t);
+        lhs_batching_axes.appendAssumeCapacity(lhs._shape.axis(l));
+        rhs_batching_axes.appendAssumeCapacity(rhs._shape.axis(r));
+    }
+
+    stdx.debug.assert(lhs._shape.dim(lhs_contracting_dim) == rhs._shape.dim(rhs_contracting_dim), "scaledDot expects contracting dimensions to be equal, got {} and {} in {f} and {f}", .{ lhs_contracting_dim, rhs_contracting_dim, lhs, rhs });
+    var lhs_contracting_axes: Axes = .empty;
```
gwenzek: this logic seems copied from dot, can you extract it in an helper function that takes tags/list of tags and return lists of batching/contrating axes ?

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+
+    const scales = amax.scale(1.0 / value_max)
+        .clamp(
+            Tensor.scalar(scale_min_normal, dt),
+            Tensor.scalar(scale_max, dt),
+        )
+        .convert(.f8e4m3fn);
+
+    const divisor = scales.convert(dt)
+        .maximum(Tensor.scalar(scale_min_normal, dt).broad(scales.shape().withDtype(dt)))
```
gwenzek: `.maximum` accepts scalar and correctly broadcast them. This would be clearer with the syntax:

`scales.convert(dt).maximum(.scalar(scale_min_normal, dt))`

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+    const scaled = if (input_global_scale) |igs|
+        x.mul(igs.convert(dt).broad(x.shape()))
+    else
+        x;
+    const grouped = scaled.splitAxis(args, .{ .sc = -1, .blk = block_size });
+    const amax = grouped.abs().max(.blk);
+
+    const scales = amax.scale(1.0 / value_max)
+        .clamp(
+            Tensor.scalar(scale_min_normal, dt),
```
gwenzek: `.clamp(.scalar(min, dt), .scalar(max,dt))`

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+        .convert(.f4e2m1)
+        .reshape(x.shape().withDtype(.f4e2m1));
+
+    return .{
+        .values = values,
+        .scales = scales.squeeze(.blk),
+    };
+}
+
+pub fn scaledDot(
```
gwenzek: scaledDot, quantize and unpack should go in `zml.nn`. Just keep `composite` here

zml/io.zig
```diff
@@ -373,7 +373,8 @@ pub const Loader = struct {
         };
 
         if (sources.len != 1) {
-            std.debug.panic("Expected loaded tensor to have only 1 source, got {}", .{sources.len});
+            log.debug("Skipping fused tensor with {} sources; expected to be loaded via loadExecute", .{sources.len});
```
gwenzek, line 376: suspect, why a debug log instead of panic ? if this result in incorrectly loaded model, this seems worse

zml/ops.zig
gwenzek: can you document the list of dtypes supported here ?

zml/ops.zig
```diff
@@ -1528,6 +1528,253 @@ pub fn customCallOutputOperandAliases(
...
+        if (lhs_tag == Shape.toTag(args)) continue;
+        if (rhs.shape().hasTag(lhs_tag)) |rhs_tag_index| {
+            batching_axes.appendAssumeCapacity(.{ @intCast(lhs_tag_index), @intCast(rhs_tag_index) });
+        }
+    }
+
+    const Axes = stdx.BoundedArray(i64, constants.MAX_RANK);
+
+    const result_dtype: DataType = switch (lhs.dtype()) {
+        .f4e2m1, .f8e4m3, .f8e4m3fn, .f8e5m2, .f8e4m3b11fnuz, .f8e4m3fnuz, .f8e5m2fnuz => .bf16,
```
gwenzek: Could we support int4 int8 kernels too ?

elogir: I will look into it but that's for later, as I'm not sure what's the state in XLA

zml/io.zig
elogir: We will have to fuse QKV using the new loader system and I'm not entirely sure how to handle the difference so this was relaxed. Although I think we will convert all of them to fused so we might end up changing that part or something else

zml/ops.zig
elogir: I'm not sure we want that in an nvfp4 function, block size must be 16 for nvfp4. Instead we could have a quantize function that is broader but should support everything else and wouldn't be typed like that

gwenzek (approved):

gwenzek:
Thanks great work!

PR #703, opened 2026-08-17
zml/pjrt: propagate ready-event failures and validate opaque pointers

neudinger:
Fix two PJRT error-handling cases that could allow execution to continue after a backend failure.

  - Query `PJRT_Event_Error` when `Event.await` receives an event that is already ready.
  - Use the same error conversion and destruction path for ready-event and callback errors.
  - Reject a successful `PJRT_Buffer_OpaqueDeviceMemoryDataPointer` result when `device_memory_ptr` is null.

### DIFF
--- a/pjrt/pjrt.zig
+++ b/pjrt/pjrt.zig
@@ -74,6 +74,13 @@ pub const ApiError = error{
     Unauthenticated,
 };
 
+inline fn interpretPjrtError(api: *const Api, pjrt_error: *Error, context: []const u8) ApiError {
+    defer pjrt_error.deinit(api);
+    const err_code = pjrt_error.getCode(api).toApiError();
+    log.err("[{s}] {t}: {s}", .{ context, err_code, pjrt_error.getMessage(api) });
+    return err_code;
+}
+
 fn InnerMixin(comptime innerT: type) type {
     return struct {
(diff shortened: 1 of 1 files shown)


neudinger:
## Why

in the custom intel pjrt plugin the error management was not correctly handled and returned nullptr (now fixed PR in process) but if a success is returned by the pjrt initialisation but a nullprt is returned zml crash with no context .

  `Event.await` previously returned success immediately when an event was already ready, without checking whether the event completed with an error.

  This could hide the original backend failure and allow execution to continue until a later operation attempted to use an invalid or null device pointer. The resulting failure was therefore disconnected from the actual PJRT error and could
  appear as a null-pointer panic.

  PJRT errors are now:

  - Logged with their original message.
  - Returned using their original PJRT error code.
  - Destroyed after being consumed.

  A non-conforming plugin result containing success with a null opaque pointer now returns `ApiError.Internal` instead of force-unwrapping the pointer.

  ## Scope

  This is limited to PJRT error propagation and pointer validation. It does not change the public ZML API or normal successful execution.

  ## Validation

  ```bash
  bazel build //pjrt:pjrt

  The modified Zig source was also checked with zig fmt --check.

hugomano (approved):

PR #706, opened 2026-08-18
platforms/oneapi:  bundle oneDNN and TBB with PJRT update PJRT_ONEAPI_RELEASE

neudinger:
platforms/oneapi: bundle oneDNN and TBB with PJRT
Add oneDNN 2026.0 and TBB 2023.1 to the hermetic oneAPI packages.
the pjrj plugin need it for the mistral models type with the support of the FP8 scaled matmul with libdnn.

platforms/oneapi: update PJRT_ONEAPI_RELEASE and SHA256 for new artifact version


neudinger:
This is required for the mistral model type to work on intel GPU

Finistere (approved):

PR #722, opened 2026-08-25
zml: mlir known types

gwenzek:
small idea I had when reading your todo. LMK what you think. Hopefully this should be slightly faster than going through the MLIR ABI for each possible candidate, but also it's not in a perf sensentive part of the code.


Corendos (approved):
Nice idea

I don't remember why I wrote that lol but as of today it's not really relevant because we already need a context to go from `DataType` to `mlir.Type` so that's symmetrical.

I'll approve but I'm not entirely sold on the "precomputed" (new version) versus "explicit" (old version)

PR #725, opened 2026-08-25
zml: Use zml.nn.Linear for forwardMoe and rename quantization structs

Finistere:
`forwardMoe` is effectively consuming a `gate_up` and `down` linear MLP, it "just" only applies parts of it.
I've renamed:
```
Quant -> Quantization
QuantSchem -> Quantization.Scheme
TensorScale -> Quantization.GlobalScale
Direction { multiplier, divisor } -> Operation { multiply, divide }
```
and moved it outside of zml.nn.Linear

This makes it easier in LLMD to just load zml.nn.Linear

### DIFF
--- a/examples/llm/models/llama/model.zig
+++ b/examples/llm/models/llama/model.zig
@@ -97,7 +97,7 @@ pub const LoadedModel = struct {
     }
 
     pub fn unloadBuffers(_: *const LoadedModel, buffers: *Buffers, allocator: std.mem.Allocator) void {
-        if (buffers.lm_head) |*lm_head| lm_head.weight.deinit();
+        if (buffers.lm_head) |*lm_head| zml.nn.Linear.unloadBuffers(lm_head);
         Llama.unloadBuffers(&buffers.model, allocator);
     }
 
@@ -189,7 +189,7 @@ pub const Model = struct {
     }
 
     pub fn unloadBuffers(self: *zml.Bufferized(Model), allocator: std.mem.Allocator) void {
-        if (self.lm_head) |*lm_head| lm_head.weight.deinit();
+        if (self.lm_head) |*lm_head| zml.nn.Linear.unloadBuffers(lm_head);
         Llama.unloadBuffers(&self.model, allocator);
     }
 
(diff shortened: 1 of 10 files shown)


examples/llm/models/llama/model.zig
```diff
@@ -480,12 +480,9 @@ const Mlp = struct {
     }
 
     pub fn unloadBuffers(self: *zml.Bufferized(Mlp)) void {
-        self.up_proj.weight.deinit();
-        if (self.up_proj.bias) |*bias| bias.deinit();
-        self.gate_proj.weight.deinit();
-        if (self.gate_proj.bias) |*bias| bias.deinit();
-        self.down_proj.weight.deinit();
-        if (self.down_proj.bias) |*bias| bias.deinit();
+        zml.nn.Linear.unloadBuffers(&self.up_proj);
```
gwenzek: `zml.Buffer.deinitAll(Mlp, self);` should work

gwenzek:
I like the improve naming with scales and global_scale, and the refactoring of loading/unloading as well as the improved moe api.

But I'm not sure I like the new QuantizedTensor 
At the core this changes:

```
pub const Linear = struct {
    weight: Tensor,
    bias: ?Tensor = null,
    tag: Shape.Tag,
    quant: ?Quantization = null,
}
```

into:

```
pub const Linear = struct {
    weight: struct { values: Tensor, quantization: ?Quantization = null },
    bias: ?Tensor = null,
    input_scale: ?QuantizedTensor.GlobalScale = null,
    tag: Shape.Tag,
}
```

I don't think the second one is better. Proof for me is that `moe` api don't even used the QuantizedTensor, only nn.Linear, showing that there is no point in separating the twos. And I feel both are too nested. 


I also don't like the multiply/divide hidden inside, feels this should be resolved at loading time, and always store the multiplicative scales.

Finistere:
> I don't think the second one is better. Proof for me is that moe api don't even used the QuantizedTensor, only nn.Linear, showing that there is no point in separating the twos. And I feel both are too nested.

Well, I agree that nothing else uses the QuantizedTensor outside of Linear and might never happen, so fair point.

gwenzek (approved):
thanks, feels like a net improvement now

zml/zml.zig
```diff
@@ -35,6 +35,7 @@ pub const CompilationOptions = module.CompilationOptions;
 pub const Compiler = module.Compiler;
 pub const moe = @import("moe/moe.zig");
 pub const nn = @import("nn.zig");
+pub const Quantization = nn.Quantization;
```
dmehala: Could we move all quantization related code into a dedicated `.zig` file? For seperation of concerns sake

zml/moe/moe.zig
```diff
@@ -189,25 +185,26 @@ pub const Metadata = union(Backend) {
...
-    bias_down: ?zml.Tensor,
-    w1_global_scale: ?zml.Tensor,
-    w2_global_scale: ?zml.Tensor,
+    gate_up: zml.nn.Linear,
+    down: zml.nn.Linear,
     opts: Options,
     metadata: Metadata,
     parameters: Parameters,
 ) !zml.Tensor {
+    try validateQuantizationPair(gate_up, down);
```
dmehala: `validateQuantizationPair` is three line, can we valid these here instead of adding an additional function?

dmehala (approved):
LGTM, much better. thanks :shipit:

PR #727, opened 2026-08-25
zml/attention: fix latent KV cache update

nzinci:
Latent `KvCache.update` currently passes one tuple to `scatterSlices`, so the branch fails to compile when instantiated.

Pass the three arguments directly and add a latent-cache compile regression test.

Tests:

- `bazel test --jobs=2 //zml:test`
- `bazel build --jobs=2 //zml`

### DIFF
--- a/zml/attention/paged_attention.zig
+++ b/zml/attention/paged_attention.zig
@@ -14,6 +14,7 @@ test {
     std.testing.refAllDecls(Backend);
     std.testing.refAllDecls(Options);
     std.testing.refAllDecls(Parameters);
+    std.testing.refAllDecls(KvCache);
 }
 
 pub const Backend = enum {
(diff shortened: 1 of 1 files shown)


zml/attention/paged_attention.zig
```diff
@@ -277,6 +277,30 @@ pub const KvCache = union(enum) {
     }
 };
 
+test "compile latent KV cache update" {
```
gwenzek: this feels overkill, the problem here is that Zig didn't even tried to compile `KvCache.update`

Can you `std.testing.refAllDecls(KvCache);` at the top of the file instead please ?

gwenzek:
thanks, this code is indeed poorly tested because we use a previous version of it internally, and we should switch to use the public version to avoid this kind of issue

dmehala (approved):
LGTM. Thanks for your contribution.

PR #729, opened 2026-08-26
zml: refactor slice API

gwenzek:
## Overview

* merge `dynamicSlice` and `slice` into `slices`
* merge `dynamicSlice1d` and `slice1d` into `slice` 
* drop `choose`, `choose1d`
* handle singleton for dynamic slices too
* typed API so that it's easier to specify arguments

## Migration guide:

```diff
// new .slice accepts all old .slice1d inputs 
- const x1 = x.slice1d(-1, .{ .start = 0, .end = half_dim });
+ const x1 = x.slice(-1, .{ .start = 0, .end = half_dim });

- k.dynamicSlice(.{ .layer = zml.Tensor.DynSlice{ .start = cache_index, .len = 1 } }).squeeze(.layer);
+ k.slice(.layer, .dynSingle(cache_index));

- k.dynamicSlice(.{
-    .layer = zml.Tensor.DynSlice{ .start = cache_index, .len = 1 }, 
- }).squeeze(.layer).choose1d(.k, kv_head);
+ k.slice(.{ .layer, .k }, &.{ .dynSingle(cache_index), .single(kv_head) });

- o.choose1d(ax, 5),
+ o.slice(ax, .single(5))
```

Now the `Tensor.Slice` struct is bigger and now as a Tensor in it, which may break user code if your model was storing a `Slice` inside it, because now `zml.Bufferize(Model)` will contain one more `?Buffer`. You can either pass an extra `null` to the `zml.Exe` or refactor to remove `Slice` out of `Model`. I haven't seen this happen so far. The idea is that `Slice` will mostly be a stack only struct, only created to call `slice`

### DIFF
--- a/examples/llm/models/lfm2/model.zig
+++ b/examples/llm/models/lfm2/model.zig
@@ -471,7 +471,7 @@ pub const ShortConv = struct {
                 const sh = tokens_position_offset.shape().insert(.last, .{ .seq = self.config.conv_L_cache });
                 break :lkp zml.Tensor.iota(sh, .seq).convert(.u32).add(left_pad.convert(.u32).broad(sh)).broad(sh);
             };
-            const scatter_data = Bx.dynamicSlice(.{ .seq = @as(zml.Tensor.DynSlice, .{ .start = start.convert(.u32), .len = @intCast(self.config.conv_L_cache) }) });
+            const scatter_data = Bx.slice(.seq, .dyn(start, self.config.conv_L_cache));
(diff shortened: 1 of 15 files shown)


tgautie:
Does this allow dynamic slices over more than one element? What would it look like?

Edit: I guess it's just passing a `.dyn` slice instead of the `.dynSingle`, from looking at the code.

tgautie (approved):
Looks great !

Finistere:
nice

PR #736, opened 2026-08-27
zml: refactor `module.zig` into `Compiler.zig`

gwenzek:
Migration instructions:

zml.module.CompilationContext -> zml.Compiler
zml.module.CompileError -> zml.Compiler.Error
zml.module.CompilationOptions -> zml.Compiler.Options
zml.module.compile -> zml.Platform.compiler or zml.Compiler.compile

Breaking changes: none, I've kept a small compatibility layer at the top level, with upgrade instructions.

We can remove it in a second time.


gwenzek:
I have more changes upcoming to the `Compiler` but wanted to do this first to simplify future code reviews

Finistere (approved):
nice

PR #737, opened 2026-08-27
zml/attention: optimize sparse MLA value accumulation

Finistere:
GLM 5.2 represents each MLA query/key head as 512 latent dimensions
followed by 64 RoPE dimensions. All 576 dimensions participate in the
attention score, but the value is only the 512-dimensional latent prefix.

Previously, sparse MLA accumulated a 576-dimensional context and sliced
off the RoPE suffix afterward. Return the 512-dimensional value context
directly instead. This also reduces the padded value accumulator from
1024 to 512 elements.

Make the value width explicit so the same implementation also supports
models such as DeepSeek V4, where the RoPE dimensions are included in
the value.

Keys and values share the same latent cache. Remove the redundant value
pointer and strides, reuse the loaded key tile for value accumulation,
and remove paging arguments that are no longer needed after converting
top-k indices to physical cache positions.

Add a grouped partial-reduction path for the high split counts that
improve occupancy on MI300, and tune the shape-based launch selection
more aggressively for GB300.

### DIFF
--- a/zml/attention/paged_attention.zig
+++ b/zml/attention/paged_attention.zig
@@ -801,8 +801,9 @@ pub fn partialSoftmax(self: zml.Tensor, axis: anytype) PartialSoftmax {
 pub const Mla = struct {
     pub const Options = struct {
         rope_rank: i64,
+        value_rank: i64,
         scale: ?f32 = null,
-        /// null selects automatically; 1 forces the 2D kernel; other values must be powers of two up to 16.
+        /// null selects automatically; 1 forces the 2D kernel; other values must be powers of two up to 128.
         num_kv_splits: ?u8 = null,
     };
 
@@ -833,9 +834,12 @@ pub const Mla = struct {
 
         const attn_weights = scores_sink.softmax(.kv);
         const attn_weights_non_sink = attn_weights.slice(-1, .{ .end = topk.dim(.topk) });
-        return attn_weights_non_sink.dot(selected_kv, .kv).convert(q.dtype());
+        const selected_values = selected_kv.slice(.hd, .{ .end = opts.value_rank });
+        return attn_weights_non_sink.dot(selected_values, .kv).convert(q.dtype());
     }
 
+    /// Computes sparse MLA scores over the complete cached key and returns
+    /// `opts.value_rank` output dimensions per head.
     pub fn pagedSparseAttention(parameters: Parameters, q: zml.Tensor, kv_cache: KvCache, sink: ?zml.Tensor, topk: zml.Tensor, tokens_pos: zml.Tensor, opts: Mla.Options) zml.Tensor {
         const latent_kv = switch (kv_cache) {
             .latent => |latent_kv| latent_kv,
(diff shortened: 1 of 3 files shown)


zml/attention/paged_attention.zig
```diff
@@ -799,11 +799,27 @@ pub fn partialSoftmax(self: zml.Tensor, axis: anytype) PartialSoftmax {
 }
 
 pub const Mla = struct {
+    pub const ValueMode = enum {
+        /// Values are the non-RoPE prefix of the cached key (GLM-style MLA).
+        latent,
+        /// Values are the complete cached key, including its RoPE suffix
+        /// (DeepSeek V4-style MLA).
+        latent_and_rope,
+    };
```
dmehala: That doesn't feel like the right abstraction. IMO, the model should pass the value rank as an option argument

zml/attention/triton_attention.zig
```diff
@@ -46,33 +46,93 @@ const SparseMlaLaunchConfig = struct {
...
+    tile_size: usize = 16,
+    max_splits: usize = 128,
+    main_num_warps: usize = 4,
+    main_num_stages: usize = 2,
+    grouped_reduce_threshold: usize = 32,
+    splits_per_group: usize = 16,
+    grouped_reduce_num_warps: usize = 1,
+    parallel_reduce_min_splits: usize = std.math.maxInt(usize),
+    parallel_reduce_num_warps: usize = 1,
+};
```
dmehala: Could we use `SparseMlaLaunchConfig` here instead? The addition of `SparseMlaTuning` seems redundant.

zml/attention/paged_attention.zig
```diff
@@ -799,11 +799,27 @@ pub fn partialSoftmax(self: zml.Tensor, axis: anytype) PartialSoftmax {
...
+
     pub const Options = struct {
         rope_rank: i64,
+        value_mode: ValueMode,
```
dmehala:
```suggestion
        value_rank: i64,
```

zml/attention/triton_attention.zig
```diff
@@ -773,55 +825,80 @@ pub const paged = struct {
                 .{
                     .cfg = kernel_config,
                     .grid = .{ @intCast(launch.direct_programs), 1, 1 },
-                    .num_warps = 4,
-                    .num_stages = 2,
+                    .num_warps = @intCast(tuning.main_num_warps),
+                    .num_stages = @intCast(tuning.main_num_stages),
```
dmehala: Shouldn’t this be `launch.*` instead? That seems to be the appropriate namespace here.

zml/attention/paged_attention.zig
Finistere: Good point

zml/attention/paged_attention.zig
Finistere: I don't know, I see your point, but it's not obvious to me that it's simpler than an enum. I've renamed it `ValueRank` rather than `ValueMode` though

Finistere: ok nevermind, did what you said, you were rigth

dmehala (approved):

PR #749, opened 2026-08-31
zml/io: record computed bindings in the store

elogir:
A binding created through `maybeCreateBinding` is computed from its sources by an executable and must never be streamed from the checkpoint. `Loader.defaultCallback` inferred that from `sources.len != 1`, which a single-source binding fails resulting to the raw bytes being loaded and producing wrong results

### DIFF
--- a/zml/io.zig
+++ b/zml/io.zig
@@ -24,8 +24,13 @@ const Tensor = @import("tensor.zig").Tensor;
 const log = std.log.scoped(.@"zml/io");
 
 pub const TensorStore = struct {
+    pub const Binding = struct {
+        tensors: []*safetensors.Tensor,
+        transformed: bool,
+    };
+
     registry: *safetensors.TensorRegistry,
-    id_to_sources: std.AutoHashMapUnmanaged(Tensor.Id, []*safetensors.Tensor),
+    id_to_sources: std.AutoHashMapUnmanaged(Tensor.Id, Binding),
     allocator: std.mem.Allocator,
     arena: std.heap.ArenaAllocator,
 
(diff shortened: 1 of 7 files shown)


zml/io.zig
```diff
@@ -338,7 +345,9 @@ pub const Loader = struct {
         progress: ?*std.Progress.Node = null,
     };
 
-    pub fn load(self: *Loader, io: std.Io, comptime T: type, model: *const T, buffers: *Bufferized(T), store: *const TensorStore, shardings: []const Sharding, opts: LoadOpts) void {
+    pub const LoadError = error{ComputedTensorNotDelivered};
+
+    pub fn load(self: *Loader, io: std.Io, comptime T: type, model: *const T, buffers: *Bufferized(T), store: *const TensorStore, shardings: []const Sharding, opts: LoadOpts) LoadError!void {
```
gwenzek, line 350: why does this return an error now ?

zml/io.zig
```diff
@@ -351,43 +360,60 @@ pub const Loader = struct {
...
-        const Ctx = struct {
-            self: *Loader,
-            io: std.Io,
-            store: *const TensorStore,
-            shardings: []const Sharding,
-            buffers: []*Buffer,
-            opts: LoadOpts,
-        };
+        const flattened_tensors = arena.allocator().alloc(*const Tensor, tensor_count) catch @panic("Errors can't be handled in `loadInner`");
+        meta.forEachVisit(model, *const Tensor, struct {
```
gwenzek: this is `zml.meta.collectBuf`

zml/io.zig
elogir: If the transform that needs to happen before it is not done, we fail so it doesn't load incorrect data

Corendos (approved):

PR #752, opened 2026-08-31
zml/ops: refactor manualComputation

dmehala:
Refactor manualComputation to make the API easier to use, less error-prone, and more readable.
As a side effect, this also adds support for nullable Tensors.


dmehala:
close in favor of #757

PR #753, opened 2026-08-31
zml/tensor: fix triangularSolve batched RHS contract

nzinci:
`Tensor.triangularSolve` currently rejects batched inputs and derives the result shape from the coefficient matrix.

Allow tensors with rank 2 or greater and preserve the right-hand side's shape in the result.

This was found while working on a separate Qwen3.5 GDN optimization.

Tests:

- `bazel test //zml:test`
- `bazel build //zml`


gwenzek (approved):
Good catch

https://github.com/openxla/stablehlo/blob/main/docs/spec.md#triangular_solve

PR #754, opened 2026-08-31
examples/llm: fix Qwen3.5 padded prefill cache

nzinci:
Qwen3.5 prefill pads prompts to the compiled sequence length. Linear attention currently treats that padded suffix as active, which changes the recurrent and convolution cache state.

Pass the active length separately from the generation position, make the padded suffix neutral for the Gated Delta Net recurrence, and build the convolution cache from the valid prefix.

Tests:

- `bazel test //examples/llm:qwen3_5_cache_tests`
- `bazel test //zml:test`
- `bazel build //examples/llm:llm //examples/llm:test`

### DIFF
--- a/examples/llm/BUILD.bazel
+++ b/examples/llm/BUILD.bazel
@@ -1,7 +1,7 @@
 load("@bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@rules_oci//oci:defs.bzl", "oci_image", "oci_load", "oci_push")
-load("@rules_zig//zig:defs.bzl", "zig_binary")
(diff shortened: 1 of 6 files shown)


gwenzek:
@tgautie can you check this one, since you're the most familiar with the model ?

tgautie:
Thanks @nzinci for the contribution, will take a look!

tgautie (changes requested):
@nzinci We won't merge this, it adds complexity to a GatedDeltaNet structure that is only showcased here as an example.

With a single sequence, the suffix issue you found can be fixed by changing the bounds of the while op, that would be a more adapted fix for this example.

Also, please respect the conventions for tests, the new file only for cache-tests in qwen3_5 is not customary.

tgautie:
Handled in https://github.com/zml/zml/pull/769

PR #756, opened 2026-09-01
pjrt: report errors on already-ready events

elogir:
PJRT_Event_Await short-circuits when the event is already ready, so a failed event returned success instead of its error. Check PJRT_Event_Error on that path too.

* pjrt/pjrt.zig (Event.await): Interpret the event error before the early return.
* zml/platform.zig (Platform.format): Iterate the devices field, not a call.

### DIFF
--- a/pjrt/pjrt.zig
+++ b/pjrt/pjrt.zig
@@ -1338,6 +1338,10 @@ pub const Event = opaque {
 
     pub fn await(self: *Event, api: *const Api, io: std.Io) ApiError!void {
         if (self.isReady(api)) {
+            if (self.getEventError(api)) |err| {
+                return interpretPjrtError(api, err, "PJRT_Event_Error");
+            }
+
             return;
         }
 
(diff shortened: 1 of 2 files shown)


pjrt/pjrt.zig
```diff
@@ -1338,6 +1338,10 @@ pub const Event = opaque {
 
     pub fn await(self: *Event, api: *const Api, io: std.Io) ApiError!void {
         if (self.isReady(api)) {
+            if (self.getEventError(api)) |err| {
```
gwenzek, line 1341: I find this a bit sus. 
If you look at XLA, `PJRT_Event_Error` calls both `IsReady` then `Await`.

so we have `if (ready) ev.await() else ev.onReady()` which is another form of `await` 

how did you trigger this error ? Is there another code path where we should check the error instead ? `awaitRaw` maybe ?

dmehala:
closing since this has been handled in #844

PR #757, opened 2026-09-01
zml/ops: refactor `manualComputation`

dmehala:
Simplify `manualComputation` by merging `inputs` and `body_context` into a single `input` argument.
Internally, collect all and shard all tensors (including `?Tensor`) from `input` before invoking
the body function.

### DIFF
--- a/zml/attention/flashattn.zig
+++ b/zml/attention/flashattn.zig
@@ -958,7 +958,7 @@ pub const paged_fa2 = struct {
                         },
                         opts: zml.ops.CustomCallOptions,
 
-                        fn body(self: @This(), _: std.mem.Allocator, output: zml.Shape) zml.Tensor {
+                        fn body(self: @This(), output: zml.Shape) zml.Tensor {
                             return zml.ops.customCall(Decode.custom_call_name, self.inputs, output, self.metadata, self.opts);
                         }
                     }).body,
@@ -1044,7 +1044,7 @@ pub const paged_fa2 = struct {
                         },
                         opts: zml.ops.CustomCallOptions,
 
-                        fn body(self: @This(), _: std.mem.Allocator, output: zml.Shape) zml.Tensor {
+                        fn body(self: @This(), output: zml.Shape) zml.Tensor {
                             return zml.ops.customCall(Prefill.custom_call_name, self.inputs, output, self.metadata, self.opts);
                         }
                     }).body,
@@ -1119,7 +1119,7 @@ pub const paged_fa2 = struct {
                         },
                         opts: zml.ops.CustomCallOptions,
 
-                        fn body(self: @This(), _: std.mem.Allocator, output: zml.Shape) zml.Tensor {
+                        fn body(self: @This(), output: zml.Shape) zml.Tensor {
                             return zml.ops.customCall(Decode.custom_call_name, self.inputs, output, self.metadata, self.opts);
                         }
                     }).body,
@@ -1596,7 +1596,7 @@ pub const paged_fa3 = struct {
                         },
                         opts: zml.ops.CustomCallOptions,
 
-                        fn body(self: @This(), _: std.mem.Allocator, output: zml.Shape) zml.Tensor {
+                        fn body(self: @This(), output: zml.Shape) zml.Tensor {
                             return zml.ops.customCall(Decode.custom_call_name, self.inputs, output, self.metadata, self.opts);
                         }
                     }).body,
@@ -1673,7 +1673,7 @@ pub const paged_fa3 = struct {
                         },
                         opts: zml.ops.CustomCallOptions,
 
-                        fn body(self: @This(), _: std.mem.Allocator, output: zml.Shape) zml.Tensor {
+                        fn body(self: @This(), output: zml.Shape) zml.Tensor {
                             return zml.ops.customCall(Prefill.custom_call_name, self.inputs, output, self.metadata, self.opts);
                         }
                     }).body,
@@ -1738,7 +1738,7 @@ pub const paged_fa3 = struct {
                         },
                         opts: zml.ops.CustomCallOptions,
 
-                        fn body(self: @This(), _: std.mem.Allocator, output: zml.Shape) zml.Tensor {
+                        fn body(self: @This(), output: zml.Shape) zml.Tensor {
                             return zml.ops.customCall(Decode.custom_call_name, self.inputs, output, self.metadata, self.opts);
                         }
                     }).body,
--- a/zml/attention/flashattn.zig
+++ b/zml/attention/flashattn.zig
@@ -947,20 +947,34 @@ pub const paged_fa2 = struct {
 
                 const output_shape = q2.shape();
                 var o = zml.ops.manualComputation(
+                    (struct {
+                        inputs: struct { zml.Tensor, zml.Tensor, zml.Tensor, zml.Tensor, zml.Tensor, zml.Tensor, zml.Tensor, zml.Tensor, zml.Tensor, zml.Tensor },
+                        metadata: struct {
+                            is_causal: bool,
+                            max_seqlen_k: usize,
+                            num_heads: i64,
+                            window_size_left: i32,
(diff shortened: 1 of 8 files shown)


zml/ops.zig
```diff
@@ -1909,6 +1909,360 @@ pub fn customCall(target_name: [:0]const u8, inputs: anytype, outputs: anytype,
...
+        input_idx: usize = 0,
+
+        fn localize(self: *@This(), _: Tensor) Tensor {
+            stdx.debug.assert(self.input_idx < self.local_tensors.len, "manualComputation ran out of localized inputs", .{});
+            defer self.input_idx += 1;
+            return self.local_tensors[self.input_idx];
+        }
+    };
+
+    var context: Context = .{ .local_tensors = local_tensors };
```
gwenzek, line 1958: this function seems not needed. We have a notion of `scope` in `Compiler.zig` that is supposed to do this. Normally you do `scope = compiler.pushBlock(block)`; and then `scope.registerTensorAsBlockArgument(tensor.id, input_idx)`

zml/tensor.zig
```diff
@@ -4590,9 +4590,9 @@ pub const Tensor = struct {
         defer ctx.arena.allocator().free(full_name);
         switch (ctx.platform.target) {
             .cpu, .cuda, .rocm, .tpu, .metal => {
-                ops.manualComputation(input, {}, full_name, (struct {
-                    fn body(_full_name: []const u8, _: std.mem.Allocator, sharded_input: Tensor, _: void) void {
-                        ops.customCall("zml$print", sharded_input, {}, .{ .name = _full_name }, .{ .has_side_effect = true });
+                ops.manualComputation(.{ .input = input, .name = full_name }, {}, (struct {
+                    fn body(body_ctx: anytype, _: std.mem.Allocator, _: void) void {
```
gwenzek: I don't really like that all `ops.manualComputation` has all arguments `anytype` but still requires some specific structure for "inputs".

Similarly now, `body_ctx` is also `anytype`, not sure why.

If you want to improve `manualComputation` I'd try to make the types more explicit rather than less.

gwenzek (commented):
at first view doesn't feel like an improvement, the API seems less typed and will not generate better error messages. What are you trying to improve ?

dmehala:
> What are you trying to improve ?

Mainly usability and readability. With the current positional API, it is easy to lose track of what each entry in `sharded_inputs` represents: 
```zig
// Extracted from `moe.zig`.
const local_num_experts = sharded_inputs[3].dim(.expert);
const partition_id = zml.ops.partitionId().convert(.i32);
...
const local_output = triton.fusedExpertsImpl(
 sharded_inputs[0],
 sharded_inputs[3],
 sharded_inputs[4],
 sharded_inputs[2],
 sharded_inputs[1],
   ...
 );
```
From this code alone, there is no clear indication of what `sharded_inputs[3]` or `sharded_inputs[4]` represent. With this change, it makes that relationship explicit:
```zig
const local_num_experts = ctx.gate_up.weight.dim(.expert);
const partition_id = zml.ops.partitionId().convert(.i32);
...
const local_output = triton.fusedExpertsImpl(
 ctx.input,
 ctx.gate_up.weight,
   ...
 );
```
This also makes the code less susceptible to positional ordering mistakes and gives us a natural way to preserve optional fields such as `?Tensor`.

> If you want to improve manualComputation I'd try to make the types more explicit rather than less.

I agree. I explored in that direction with #752 but it has it own flaw: it introduced an implicit interface that every manual computation context had to satisfy.

All in all, I came to the conclusion that there's no silver bullet (or I'm not good enough to find one) and went with the less disruptive (API usage wise) solution. If you have suggestion to improve or make it better I am more than happy to read you suggestions.

EDIT: It can also lead to a **saner** usage, as below:
```zig
const MoE = struct {
  input: Tensor,
  topk_ids: Tensor,
  gate_up: nn.Linear,

  pub fn call(self: @This(), allocator: std.mem.Allocator, output_shape: Shape) Tensor {
     const local_num_experts = self.gate_up.weight.dim(.expert);
      ....
  }
};

manualComputation(
  MoE{
    .input = input,
    .topk_ids = topk_ids,
  },
  output_shape,
  MoE.call,
);
```

zml/attention/flashattn.zig
```diff
@@ -947,20 +947,34 @@ pub const paged_fa2 = struct {
 
                 const output_shape = q2.shape();
                 var o = zml.ops.manualComputation(
+                    (struct {
```
Finistere: nitpick: would be nice to convert this to actually use a struct rather than an array of zml.Tensor

zml/ops.zig
```diff
@@ -1959,14 +2315,14 @@ pub fn manualComputation(
...
-    const sharded_outputs: []const Tensor = manualComputationInternal(inputs_, output_shapes, body_context, body_fn) catch |err| switch (err) {
+    const sharded_outputs: []const Tensor = manualComputationInternalLegacy(inputs_, output_shapes, body_context, body_fn) catch |err| switch (err) {
         error.OutOfMemory => @panic("OOM"),
     };
     const ReturnT = manualComputationReturnType(body_fn);
     return manualComputationSliceToReturn(ReturnT, sharded_outputs);
 }
 
-fn manualComputationInternal(
+fn manualComputationInternalLegacy(
```
gwenzek: please remove the legacy variant

zml/ops.zig
```diff
@@ -1909,6 +1909,362 @@ pub fn customCall(target_name: [:0]const u8, inputs: anytype, outputs: anytype,
...
+        @intFromPtr(lhs.value()),
+        @intFromPtr(rhs.value()),
+    };
+
+    const NestedStruct = struct {
+        nested_inputs: NestedInputs,
+        slice_inputs: []const Tensor,
+        original_values: [5]usize,
+
+        pub fn compute(self: @This(), _: std.mem.Allocator, output_shape: Shape) Tensor {
```
gwenzek: can you remove the allocator ? 
if a user needs an allocator, `zml.Compiler.current().arena.allocator()` works

zml/ops.zig
dmehala: addressed in 785846030bc78a2a6dbadf15bc94540db8512f8c

zml/ops.zig
dmehala: addressed in 6b2047631917651932e39c31fc36efeeb28aa1ad

hugomano (approved):
The API looks better right now

zml/attention/triton_attention.zig
```diff
@@ -975,64 +984,70 @@ pub const paged = struct {
...
-                        .query_start_len = sharded_inputs[7],
-                        .options_ = ctx_.options,
+                        .block_table = self.block_table,
+                        .seq_lens = self.seq_lens,
+                        .query_start_len = self.query_start_len,
+                        .options_ = self.options,
                     };
 
-                    const topk_final = topkToPhysical(parameters_, sharded_inputs[3], sharded_inputs[4], block_size);
-                    const active_query_count = sharded_inputs[7]
```
gwenzek, line 1008: goodbye `shareded_inputs[7]` you will be missed

zml/ops.zig
```diff
@@ -1959,22 +1936,39 @@ pub fn manualComputation(
...
+        fn localize(self: *@This(), _: Tensor) Tensor {
+            stdx.debug.assert(self.input_idx < self.local_tensors.len, "manualComputation ran out of localized inputs", .{});
+            defer self.input_idx += 1;
+            return self.local_tensors[self.input_idx];
+        }
+    };
+
+    var context: Context = .{ .local_tensors = local_tensors };
+    var local_inputs: @TypeOf(inputs) = undefined;
+    meta.mapAlloc(Context.localize, allocator, &context, inputs, &local_inputs) catch unreachable;
```
gwenzek: `catch @panic("OOM")` 

"unreachable" should be for stuff that is proven to be unreachable

gwenzek (approved):

PR #758, opened 2026-09-01
zml/safetensors: close model discovery handles

nzinci:
Model discovery leaves its safetensors entrypoint and repository handles open after building the tensor registry, including when parsing fails.

Close handles owned by `fromRepo` and `fromPath`, and close repositories at the existing `fromRepo` call sites.


gwenzek (approved):
thanks

PR #762, opened 2026-09-02
examples/llm: project only final LFM2 prefill token

nzinci:
LFM2 prefill projected every compiled sequence position through the vocabulary head, although generation uses only the last valid prompt position. This change projects and samples only `actual_seq_len - 1`. Decode is unchanged, including top-k RNG advancement.

Validated with `LiquidAI/LFM2.5-1.2B-Instruct` in bf16 on an RTX 4090 using PJRT CUDA and FlashAttention 2. Greedy and top-k runs at compiled lengths 128, 512, and 2048 produced identical selected tokens, RNG states, and next 8 tokens.

Median prefill latency with a 66-token prompt and top-k 4:

- 128: 6.13 ms to 5.79 ms, 5.6% faster
- 512: 10.77 ms to 9.69 ms, 10.0% faster
- 2048: 35.74 ms to 30.17 ms, 15.6% faster

At length 2048, median LM-head time fell from 5.78 ms to 0.41 ms.

This may also help other supported and future models. I considered a reusable module, but each model connects sampling, token-buffer updates, and prefill/decode differently, so applying the optimization within each model's inference path is simpler.

### DIFF
--- a/examples/llm/models/lfm2/inference.zig
+++ b/examples/llm/models/lfm2/inference.zig
@@ -96,11 +96,69 @@ pub const CompiledModel = struct {
 
 pub const Inference = CompiledModel;
 
+const SamplingHead = struct {
+    pub const Input = struct {
+        lm_head: model.LmHead,
+        embed_tokens: model.TokenEmbedding,
+        hidden: zml.Tensor,
+        tokens: zml.Tensor,
+        actual_seq_len: zml.Tensor,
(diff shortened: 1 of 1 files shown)


gwenzek:
Thanks for the PR. I don't think this is the right place to handle it, it should be done by the LmHead. Also the issue is pervasive across all models implementation, that's on us to fix.

PR #763, opened 2026-09-03
attention/flashattn: fix sliding window off-by-one error

tgautie:
We fix an off-by-one error in Flash-Attn: `window_size_left` is inclusive of the current position, so it is minus 1 compared to the actual `sliding_window` size.

Ref: https://github.com/Dao-AILab/flash-attention#how-to-use-flashattention


Corendos (approved):

gwenzek:
I'm surprised that the attention tests aren't broken by this PR

gwenzek:
can you also remove the `std.log.warn` in `zml/slice.zig` ? I left it by mistake

tgautie:
> I'm surprised that the attention tests aren't broken by this PR

The test was covering only the disabled -1 case, I'm making the test cover this

PR #764, opened 2026-09-03
zml/Compiler: Provide accurate device memory size to XLA

Finistere:
By default, XLA assumes 80% of the memory to be available during compilation. It wasn't an issue as long as we had one executable per layer, because most of the weights weren't parts of the inputs, but for a full forward executable we exceed that 80% threshold in llmd with the cache being sized to use almost everything up to the BFC limit. So XLA ends up generating an executable the least possible temporary buffers even though it has more than that.

### DIFF
--- a/pjrt/pjrt.zig
+++ b/pjrt/pjrt.zig
@@ -77,7 +77,7 @@ pub const ApiError = error{
 inline fn interpretPjrtError(api: *const Api, pjrt_error: *Error, context: []const u8) ApiError {
     defer pjrt_error.deinit(api);
     const err_code = pjrt_error.getCode(api).toApiError();
-    log.err("[{s}] {t}: {s}", .{ context, err_code, pjrt_error.getMessage(api) });
+    log.warn("[{s}] {t}: {s}", .{ context, err_code, pjrt_error.getMessage(api) });
     return err_code;
 }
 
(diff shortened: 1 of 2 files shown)


pjrt/pjrt.zig
```diff
@@ -77,7 +77,7 @@ pub const ApiError = error{
 inline fn interpretPjrtError(api: *const Api, pjrt_error: *Error, context: []const u8) ApiError {
     defer pjrt_error.deinit(api);
     const err_code = pjrt_error.getCode(api).toApiError();
-    log.err("[{s}] {t}: {s}", .{ context, err_code, pjrt_error.getMessage(api) });
+    log.warn("[{s}] {t}: {s}", .{ context, err_code, pjrt_error.getMessage(api) });
```
dmehala: is this expected?

gwenzek (approved):

zml/Compiler.zig
```diff
@@ -686,6 +686,26 @@ fn compileModuleToPjrtExecutable(arena: std.mem.Allocator, io: std.Io, platform:
...
+            var device_memory_size: ?u64 = null;
+            for (platform.devices) |device| {
+                const bytes_limit = device.memoryStats().bytes_limit orelse {
+                    device_memory_size = null;
+                    break;
+                };
+                device_memory_size = @min(device_memory_size orelse bytes_limit, bytes_limit);
+            }
+            if (device_memory_size) |bytes_limit| {
+                c.xla_ExecutableBuildOptionsProto_set_device_memory_size(
```
gwenzek: this fits on one line, no need to over indent

PR #769, opened 2026-09-04
nn/gated_delta_net: bound GDN recurrence

tgautie:
Fixing an issue raised in https://github.com/zml/zml/pull/754, to bound the GDN recurrence in the Qwen3.5 example model to the number of active tokens.

We also make qwen3.5 and qwen3.5-moe slightly more consistent.


gwenzek (approved):

tgautie:
@nzinci FYI handling it here

PR #770, opened 2026-09-04
stdx/json: {Union,TaggedUnion}Helpers

gwenzek:
Deprecated version:

```zig
const Answer = union {
   number: i32,
   numbers: []const i32,
};

const Message = struct {
   question: []const u8;
   answer: stdx.json.Union(Answer);
}
```

Recommended version:

```zig
const Answer = union {
    number: i32,
    numbers: []const i32,
    const Helpers = stdx.json.UnionHelpers(@This());
    pub const jsonParse = Helpers.jsonParse;
    pub const jsonParseFromValue = Helpers.jsonParseFromValue;
    pub const jsonStringify = Helpers.jsonStringify;
};

const Message = struct {
   question: []const u8;
   answer: Answer;
}
```

The new version is more verbose (but more explicit) in the struct declarion,
but actually **less** verbose at call site because it doesn't introduce an intermediary 
struct with a single `value` field.

### DIFF
--- a/examples/llm/models/llama/model.zig
+++ b/examples/llm/models/llama/model.zig
@@ -10,10 +10,7 @@ const log = std.log.scoped(.llama);
 
 pub const Config = struct {
     bos_token_id: u32,
-    eos_token_id: stdx.json.Union(union(enum) {
-        int: u32,
-        ints: []u32,
-    }),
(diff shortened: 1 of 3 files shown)


Corendos (approved):
Nice

PR #777, opened 2026-09-07
zml/platform: one ComputeCapability for every CUDA gate

elogir:
Every CUDA gate used to parse the compute capability string itself. They now share `platform.cuda.computeCapability`, which returns null off CUDA and offers `.is` / `.atLeast` / `.sm`.

Also folds the NVFP4 scale clamp into its reduce to remove one launch per projection, and is equivalent.

### DIFF
--- a/zml/attention.zig
+++ b/zml/attention.zig
@@ -25,16 +25,8 @@ pub const Backend = enum {
     pub fn auto(platform: *const zml.Platform) Backend {
         return switch (platform.target) {
             .cuda => b: {
-                const first_device = platform.pjrt_client.devices(platform.pjrt_api)[0];
-
-                if (zml.platform.cuda.tryGetComputeCapabilities(platform, first_device)) |cc| {
-                    break :b if (std.mem.eql(u8, cc, "9.0"))
-                        .cuda_fa3
-                    else
-                        .cuda_fa2;
-                }
-
-                break :b .vanilla;
+                const cc = zml.platform.cuda.computeCapability(platform) orelse break :b .vanilla;
+                break :b if (cc.is(9, 0)) .cuda_fa3 else .cuda_fa2;
             },
             .neuron => .nki,
             .metal => .metal_fa,
@@ -49,12 +41,7 @@ pub const Backend = enum {
             .nki => platform.target == .neuron,
             .metal_fa => platform.target == .metal,
             .cuda_fa2 => platform.target == .cuda,
-            .cuda_fa3 => {
-                if (platform.target != .cuda) return false;
-                const first_device = platform.pjrt_client.devices(platform.pjrt_api)[0];
-                const cc = zml.platform.cuda.tryGetComputeCapabilities(platform, first_device) orelse return false;
-                return std.mem.eql(u8, cc, "9.0");
-            },
+            .cuda_fa3 => (zml.platform.cuda.computeCapability(platform) orelse return false).is(9, 0),
         };
     }
 };
--- a/zml/attention.zig
+++ b/zml/attention.zig
@@ -26,7 +26,7 @@ pub const Backend = enum {
         return switch (platform.target) {
             .cuda => b: {
                 const cc = zml.platform.cuda.computeCapability(platform) orelse break :b .vanilla;
-                break :b if (cc.is(9, 0)) .cuda_fa3 else .cuda_fa2;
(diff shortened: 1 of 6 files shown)


zml/platform.zig
```diff
@@ -795,14 +795,40 @@ pub const CreateOptions = struct {
...
+
+        pub fn parse(text: []const u8) ?ComputeCapability {
+            var parts = std.mem.splitScalar(u8, text, '.');
+            return .{
+                .major = std.fmt.parseInt(u8, parts.first(), 10) catch return null,
+                .minor = std.fmt.parseInt(u8, parts.next() orelse "0", 10) catch return null,
+            };
+        }
+
+        pub fn is(self: ComputeCapability, major: u8, minor: u8) bool {
```
dmehala: _nit_: I'd suggest using eql here to better align with Zig std idioms.

Also, accepting a struct would make this API less error-prone. With the current signature, it's easy to accidentally swap the `major` and `minor` arguments, and the compiler can't help catch that mistake.

Could we instead accept a struct, e.g.:

```zig
// usage:
computeCapability(platform).eql(.{ .major = 9, .minor = 0 });

// instead of:
computeCapability(platform).eql(9, 0);
```

zml/quantization.zig
```diff
@@ -145,13 +147,7 @@ pub fn quantizeNvfp4(x: Tensor, input_global_scale: ?Tensor, axis: anytype) Quan
...
 
 fn supportsNvfp4InputQuantization(platform: *const Platform) bool {
-    if (platform.target != .cuda) return false;
-
-    const device = platform.pjrt_client.devices(platform.pjrt_api)[0];
-    const capability = platform_mod.cuda.tryGetComputeCapabilities(platform, device) orelse return false;
-    const major = std.fmt.parseInt(u8, std.mem.sliceTo(capability, '.'), 10) catch return false;
-
-    return major >= 10;
+    return (platform_mod.cuda.computeCapability(platform) orelse return false).atLeast(10, 0);
```
dmehala: _nit_: it doesn't read fluently, how about:
```suggestion
    return if (platform_mod.cuda.computeCapability(platform)) |cc| cc.atLeast(10, 0) else false;
```

zml/platform.zig
```diff
@@ -795,14 +795,40 @@ pub const CreateOptions = struct {
...
+        major: u8,
+        minor: u8,
+
+        pub fn parse(text: []const u8) ?ComputeCapability {
+            var parts = std.mem.splitScalar(u8, text, '.');
+            return .{
+                .major = std.fmt.parseInt(u8, parts.first(), 10) catch return null,
+                .minor = std.fmt.parseInt(u8, parts.next() orelse "0", 10) catch return null,
+            };
+        }
```
dmehala, line 808: we should be able to use [std.SemanticVersion](https://ziglang.org/documentation/0.16.0/std/#std.SemanticVersion) instead, right?

zml/attention.zig
```diff
@@ -49,12 +41,7 @@ pub const Backend = enum {
             .nki => platform.target == .neuron,
             .metal_fa => platform.target == .metal,
             .cuda_fa2 => platform.target == .cuda,
-            .cuda_fa3 => {
-                if (platform.target != .cuda) return false;
-                const first_device = platform.pjrt_client.devices(platform.pjrt_api)[0];
-                const cc = zml.platform.cuda.tryGetComputeCapabilities(platform, first_device) orelse return false;
-                return std.mem.eql(u8, cc, "9.0");
-            },
+            .cuda_fa3 => (zml.platform.cuda.computeCapability(platform) orelse return false).is(9, 0),
```
dmehala: _nit_: This reads oddly to me. At the end of the day, wouldn't we want something more like:

```zig
zml.platform.cuda.computeCapabilityEql(platform, .{ .major = 9, .minor = 0 });
```

This also makes the intent at the call site clearer.

dmehala (approved):
I left mostly nits that would improve readability/expressiveness. Nice improvements, thank you.

zml/attention.zig
elogir: Not sure about that for now since we now do `(platform_mod.cuda.computeCapability(platform)) |cc|`

zml/platform.zig
elogir: `std.SemanticVersion` forces a patch version number

Corendos (approved):

PR #786, opened 2026-09-09
zml/moe: enforce `padded_num_experts` as a power of two

dmehala:
Fix an issue when `padded_num_experts` computed doesn't land as a power of two. Notably for DeepSeek-v4 Pro with its 384 experts, divided by 32 warp is not of power of two, leading to a crash in the triton which expects a power of two number of experts.

### DIFF
--- a/zml/moe/triton.zig
+++ b/zml/moe/triton.zig
@@ -416,8 +416,7 @@ fn alignBlockSize(topk_ids: Tensor, num_experts: i64, block_size_m: i64) struct 
     else
         num_assignments + num_experts * (block_size_m - 1);
     const max_num_m_blocks = std.math.divCeil(i64, max_num_tokens_padded, block_size_m) catch unreachable;
-    const warp_size: i64 = 32;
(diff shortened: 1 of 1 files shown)


zml/moe/triton.zig
```diff
@@ -417,7 +417,7 @@ fn alignBlockSize(topk_ids: Tensor, num_experts: i64, block_size_m: i64) struct
         num_assignments + num_experts * (block_size_m - 1);
     const max_num_m_blocks = std.math.divCeil(i64, max_num_tokens_padded, block_size_m) catch unreachable;
     const warp_size: i64 = 32;
-    const padded_num_experts = (std.math.divCeil(i64, num_experts, warp_size) catch unreachable) * warp_size;
```
Finistere, line 420: why max warp_size?

loupicaaa: It is a needed parameter for the triton kernel MoeAlignBlockSize, it aligns the token in front of the experts and needs the hardware features to make the moe fast then.

loupicaaa (approved):

PR #791, opened 2026-09-10
zml/attention: Unified attention cutile backend.

loupicaaa:
This PR add a version of the attention using the zig cutile dsl in zml. The kernel is tested and made for blackwell (this could be extended to tother compute capability). As a reference, it uses this implementation [Nvidia fmha blackwell implementation](https://github.com/NVIDIA/TensorRT-Edge-LLM/blob/main/kernelSrcs/fmha_cutedsl_blackwell/fmha.py)

### DIFF
--- a/zml/attention/cutile_attention.zig
+++ b/zml/attention/cutile_attention.zig
@@ -1,0 +1,547 @@
+const std = @import("std");
+const zml = @import("../zml.zig");
+const cut = zml.kernel.cuda_tile;
+const triton = @import("triton_attention.zig");
+const AttentionOptions = @import("paged_attention.zig").AttentionOptions;
+
+pub const Options = triton.paged.Options;
+pub const Parameters = triton.paged.Parameters;
+
+pub const Config = struct {
+    batch: i64,
+    // Query storage capacity can differ from the number of metadata slots.
+    query_tokens: i64,
+    heads: i64,
+    pages: i64,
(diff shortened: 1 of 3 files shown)


zml/attention/cutile_attention.zig
```diff
@@ -0,0 +1,916 @@
...
+    strides: [4]i64,
+    head_dim: i64 = 128,
+    group_size: i64 = 4,
+};
+
+/// Explicitly list the Blackwell targets supported by the kernel (will have to be extended for future architectures)
+pub fn isBlackwell(major: u32, minor: u32) bool {
+    return (major == 10 and (minor == 0 or minor == 3)) or
+        (major == 11 and minor == 0) or (major == 12 and minor <= 1);
+}
```
dmehala: not sure this code should be here. instead in `zml.cuda...`? also there's a new api `.atLeast`

dmehala, line 30: not sure this code should live here. maybe `zml.cuda...` is better? depends is its reused anywhere else in the codebase. Probably `zml.platform.cuda.isBlackwell(platform)`?

It would be more correct to use:
```zig
pub fn isBlackwell(cc: cuda.ComputeCompatibility) bool {
   return cc.atLeast(.{ .major = 10, .minor = 3 });
}
```

dmehala (approved):
LGTM.

PR #797, opened 2026-09-10
zml/triton: wire TMA and global scratch memory size in Triton kernels

Corendos:
XLA supports TMA and scratch memory in the triton custom call but we never wired it.

It is now solved.


elogir (approved):
Nice

PR #801, opened 2026-09-11
zml/moe: add `triton_mxfp4` backend

Corendos:
This is a kernel that is quite specific to Deepseek V4 Flash

Ideally, we would want to unify the triton kernels.


Finistere (approved):

zml/moe/triton_mxfp4.zig
```diff
@@ -0,0 +1,123 @@
...
+        const hidden = self.w2.dim(1);
+        const intermediate = self.w2.dim(2) * 2;
+        const tokens: i64 = @intCast(self.input.count() / @as(usize, @intCast(hidden)));
+
+        var ids = self.ids.convert(.i32);
+        if (self.expert_parallel) {
+            const partition_id = zml.ops.partitionId().convert(.i32);
+            ids = ids.sub(partition_id.scale(experts));
+        }
+
```
dmehala, line 96: the kernel enforces that scales `s1` and `s2` should be divisble by `16`. If that's not the case we should pad. It can happend with DeepSeek v4.1 for the w2 scale. Since the intermediary dimension is `2304`. `dim = 2304/32=72` and `72%16 != 0`.

Corendos: Addressed in https://github.com/zml/zml/pull/801/commits/4dca5e83c59bbd25eb1ec39a2994f016f40284d7

PR #807, opened 2026-09-14
Gw/data parallel

gwenzek:


zml/io.zig
```diff
@@ -1381,58 +1381,6 @@ pub const DirectMemoryWriter = struct {
     }
 };
 
-fn buildMesh2x2(
```
gwenzek, line 1421: this was useful for testing so I promoted it to `zml.Sharding`

zml/ops.zig
```diff
@@ -2633,6 +2684,7 @@ test "manualComputation handler API" {
             .rhs = rhs,
         },
         shape,
+        .{},
```
gwenzek, line 2822: I don't really understand this test, nor how it can pass without specifying a partitioning dimension

@dmehala

zml/Sharding.zig
```diff
@@ -1215,47 +1252,44 @@ pub const LogicalMesh = struct {
...
 pub const AxisList = stdx.BoundedArray(PhysicalAxisTag, MAX_MESH_RANK);
+
+/// Index into one Data object's resolved mesh, not a physical hardware tag.
+pub const ViewAxisId = enum(u8) { _ };
+pub const ViewAxes = stdx.BoundedArray(ViewAxisId, MAX_MESH_RANK);
+
 pub const Binding = struct {
     logical: Shape.Tag,
-    physical: AxisList,
+    axes: ViewAxes = .empty,
```
gwenzek, line 1283: this the big change in this PR.

before each logical axis would map to a list of physical axis.
now they map to a list of "view axes".

Each "view axis" can be resolved to a `ResolvedAxis` that can map to one, or more physical axis, or to a slice of physical axis.

gwenzek: this file has been clanked, it seems "correct" but still find it a bit too complicated for my taste. Maybe I need to think of a better way of describing sharded layouts

According to the tests cases the generated MLIR hasn't changed.
I'll come back to it later

PR #824, opened 2026-09-17
ci: use bazel --experimental_disk_cache_gc_max_size

gwenzek:


cerisier:
This won't have any effect I think because GC triggers only after idle time of the bazel server. not suitable for CI one off jobs.

cerisier:
if you have long living machines then it’s worth it tho.

gwenzek:
yes, this is for the CI that "runs-on" the 4090 in the office.

oboulant (approved):

PR #836, opened 2026-09-18
zml/testing: fix environment init

Corendos:


dmehala (approved):

zml/testing.zig
```diff
@@ -14,11 +14,15 @@ var _platform: ?*const Platform = null;
...
     if (_platform == null) {
+        const memory_fraction = 0.90;
         const platform = Platform.auto(
             std.heap.c_allocator,
             std.testing.io,
             .{
-                .xla_gpu = .{ .allocator = .{ .bfc = .{ .preallocate = false, .memory_fraction = 0.85 } } },
+                .cuda = .{ .allocator = .{ .vmm = .{ .memory_fraction = memory_fraction } } },
+                .rocm = .{ .allocator = .{ .vmm = .{ .memory_fraction = memory_fraction } } },
+                .metal = .{ .allocator = .{ .bfc = .{ .preallocate = true, .memory_fraction = memory_fraction } } },
```
gwenzek, line 24: `preallocate` should be set to `false` otherwise it can trigger OOM when two users run tests on the same machine, even though the unit tests use a small amount of device memory

